PR #12, opened 2015-07-28
Print a better error message on bad_trap()

palmer-dabbelt:
This prints out the name of the trap (so I don't have to look them up)
along with mbadaddr when relevant.


aswaterman:
This is a good idea, but it should be implemented using the DECLARE_CAUSE macro to avoid code duplication.  (I'm guessing you didn't use it because it was broken, but it should now do the trick.)

PR #16, opened 2016-02-28
If atomics were disabled then emulate even in asm

colinschmidt:
There is one use of atomics in the assembly code for the pk. The other uses of atomics are in C and use this define to avoid calling using native atomics and instead emulate them. This commit emulates the assembly use of an atomic, non-atomically. This won't reliably work with multiple hearts but I don't think there is an easy way around that without atomics.

### DIFF
--- a/pk/mentry.S
+++ b/pk/mentry.S
@@ -179,6 +179,9 @@ mentry:
   j init_first_hart
 
 .LmultiHart:
+#ifndef PK_ENABLE_ATOMICS
+  j .LmultiHart
+#else
   # make sure our hart id is within a valid range
   li a1, MAX_HARTS
   bgeu a0, a1, .LmultiHart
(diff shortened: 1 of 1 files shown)


aswaterman:
We can't really support multiple harts without atomics... How about making this fail more obviously?

    #ifndef PK_ENABLE_ATOMICS
      j .LmultiHart
    #else
      # make sure our hart id is within a valid range
      li a1, MAX_HARTS
      bgeu a0, a1, .LmultiHart

      # signal we're ready to boot
      csrw mscratch, x0
      fence
      li a1, 1
      sll a1, a1, a0
      la a2, booted_harts_mask
      amoor.w x0, a1, (a2)

      # wait for main hart to grant us a stack
    1:csrr sp, mscratch
      beqz sp, 1b

      j init_other_hart
    #endif

aswaterman:
BTW, it looks like I already fixed it this way on the priv-1.9 branch.

colinschmidt:
That makes more sense. Thanks for the suggestion.

I continued to use the PK_ENABLE_ATOMICS (not __riscv_atomic) since thats what the other atomics are based on.

PR #17, opened 2016-03-22
fix multilib build error. it's type casting related.

zizztux:
I fixed a typo. It causes compile error for riscv32.


aswaterman:
thanks

PR #22, opened 2016-06-10
initialize num_harts to 0 when query harts

wsong83:
Seems num_harts is not initialized by anyone. It shows up a random value in my actual FPGA run but OK in simulation.

### DIFF
--- a/machine/configstring.c
+++ b/machine/configstring.c
@@ -60,6 +60,7 @@ static void query_hart_plic(const char* config_string, hls_t* hls, int core, int
 
 static void query_harts(const char* config_string)
 {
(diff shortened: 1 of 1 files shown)


aswaterman:
It's in .bss so should start out as zero.  Does your loader not initialize the .bss section?  What about .data?

If your loader initializes .data and not .bss, I think the right thing to do is to memset the whole .bss section.  num_harts is not the only variable that expects its value to be initially zero.

If your loader doesn't initialize .data, there's a bigger problem.

wsong83:
Thanks! There must be something messed up then.

PR #23, opened 2016-06-27
Update config.guess and config.sub

lluixhi:
I had some trouble building the proxy kernel with a musl toolchain (host triple: riscv64-pc-linux-musl), and replacing the config.guess/config.sub fixed the issue.


palmer-dabbelt:
This diff is super big.  Is this just because we haven't updated gnuconfig in a long time?  Does your update differ from upstream <http://git.savannah.gnu.org/cgit/config.git/tree/>?

On a somewhat unrelated note (assuming this is just a bump to upstream gnuconfig), I think "riscv*-pc*" is probably a bogus tuple: we're not building PCs.  It might go through, as "*-pc-linux-*" is usually meaningless, but we've been using "riscv{32,64}-unknown-{linux-gnu,elf}", so I assume you want "riscv64-unknown-linux-musl" here?

lluixhi:
@palmer-dabbelt 
Yes, this is because the current gnuconfig is out of date by about ~6 years. No, these versions are the same as at the current trunk of upstream.

Whoops. Yes. It's supposed to be `riscv64-unknown-linux-musl`

palmer-dabbelt:
Thanks, but this is out of date and appears to have a conflict.  I'm assuming some other PR updated the config stuff.

PR #24, opened 2016-06-28
Add Hwacha support.

lluixhi:
This is mostly stuff ripped out of esp-tests.

It appears to work, but as it seems there are other bugs in my test code, I'm not 100% sure.

### DIFF
--- a/config.h.in
+++ b/config.h.in
@@ -33,6 +33,9 @@
 /* Define if floating-point emulation is enabled */
 #undef PK_ENABLE_FP_EMULATION
 
+/* Define if Hwacha support is enabled */
+#undef PK_ENABLE_HWACHA
+
 /* Define if the RISC-V logo is to be displayed */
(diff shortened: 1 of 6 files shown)


aswaterman:
Since Hwacha is a non-standard extension, this should be maintained on a branch rather than merged into pk.

lluixhi:
@aswaterman 
Will github allow me to create pull requests for a branch that does not yet exist?
Or will you merge these changes into a `hwacha` branch?

aswaterman:
I'd like @colinschmidt to weigh in on how to maintain it, since he's in charge of esp-tools.

palmer-dabbelt:
Thanks, but Hwacha is somewhat defunct and the tools are not maintained.

colinschmidt:
Sorry was on vacation. I don't think it makes sense to upstream this.

PR #27, opened 2016-07-16
Support 32bit build

pmundkur:
Add some configure support for 32-bit build, instead of manually editing the default 64-bit makefile.  This will require adjusting the build.sh script of riscv-tools to remove the install subdir from the --prefix argument for the pk build.


aswaterman:
thanks.  note this broke build.sh in riscv-tools (as riscv64-unknown-elf would get appended twice to the install path) but i pushed a fix to that, too.

pmundkur:
Thanks.  riscv-pk/README.md also needs fixing; I missed that.

PR #31, opened 2016-08-09
Ignore the unexpected htif interrupt

bukinr:
First core writes a char and waits for putc event, second core should ignore the putc event.
this fixes operation with FreeBSD


palmer-dabbelt:
Sorry, but HTIF is somewhat defunct, this PR has some conflicts, and has a bunch of merges.

PR #37, opened 2016-09-09
machine, emulation.c: fix the condition of rdtime emulation

sashimi-yzh:
The time counter is enabled with the bit field set in `counteren` CSR.


aswaterman:
oops, thanks.

PR #38, opened 2016-09-15
machine, mentry.S: fix not receiving IPI for other harts

sashimi-yzh:
* Before waiting for IPI by executing wfi instruction, MSIE bit
  should be set in mie CSR for other harts, else they will get
  stuck at the wfi instruction.


aswaterman:
Good point.  Note you can just use a csrwi instruction here.

PR #40, opened 2016-09-21
Make sure pages for ELF sections have correct protection

zhemao:
I'm not sure whether this is necessary or not, but pk/bbl hasn't been correctly protecting the sections of the ELF it's loading. This could help debug some issues like code accidentally modifying the .rodata or .text sections or jumping outside of the .text section.

### DIFF
--- a/pk/elf.c
+++ b/pk/elf.c
@@ -10,6 +10,19 @@
 #include <elf.h>
 #include <string.h>
 
+/**
+ * The protection flags are in the p_flags section of the program header.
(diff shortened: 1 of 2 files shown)


aswaterman:
Could you add macros to elf.h for PF_X etc., and use the proper mmap flags PROT_EXEC etc., rather than magic constants?

zhemao:
I've made the changes. Is this what you meant?

aswaterman:
yep, thanks.  i'm quite curious to see what this will break...

PR #49, opened 2017-03-18
Elide a GCC warning

palmer-dabbelt:
We're killing config string anyway, right?


aswaterman:
This is already dead on the priv-1.10 branch.

palmer-dabbelt:
So should I merge it?  As it stands, pk doesn't build with riscv-gcc/riscv-next, and I don't want to bump submodule pointers with stuff breaking.

aswaterman:
sure

PR #51, opened 2017-03-29
Allow appending to LIBS when running make

seldridge:
Like `CFLAGS` and `LDFLAGS`, allow the user to append to `LIBS` when running `make`.

I have a usage case where I'm adding some system calls to the Proxy Kernel and I need to link against some other libraries. While I'm already modifying the Proxy Kernel with wrappers to the code for these systemcalls it would be nice if I can specify the library linkages on the command line without modifying the build flow.


palmer-dabbelt:
Sorry, but this PR has some conflicts.  I'm afraid I understand very little of automake so I'm going to close it, but feel free to re-open it with a conflict-free version if it makes sense.

PR #55, opened 2017-05-25
Add sbi-disk mcall routines

zhemao:


aswaterman:
Please maintain this on a branch; we don't want to put more abstract devices into the SBI.

zhemao:
Okay

PR #58, opened 2017-08-02
Add a platform interface

palmer-dabbelt:
Right now we can't boot exactly the same BBL on spike and the FPGA.  We'd been maintaining a fork at http://github.com/sifive/riscv-pk, but this got very out of date.  In order to avoid maintain a fork I've decided to add a "platform" behind which these details live.  I'm hoping to kill this when the platform spec is good enough so I haven't done any actual work to design the interface, but for now I think it's good enough to avoid forking again.

This adds "--with-platform", a new autoconf switch, that lets users select the platform to target.  There's two right now: spike (the default) and sifive-vc707-devkit (which boots on our FPGA).

### DIFF
--- a/bbl/logo.c
+++ b/bbl/logo.c
@@ -1,32 +1,8 @@
 #include <string.h>
 #include "mtrap.h"
-
-static const char logo[] =
-"              vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv\n"
-"                  vvvvvvvvvvvvvvvvvvvvvvvvvvvv\n"
-"rrrrrrrrrrrrr       vvvvvvvvvvvvvvvvvvvvvvvvvv\n"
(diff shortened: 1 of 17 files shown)


bbl/logo.c
```diff
@@ -1,32 +1,8 @@
...
-"rrrrrrrrrrrrrrrrrrrr      rrrrrrrrrrrrrrrrrrrr\n"
-"rrrrrrrrrrrrrrrrrrrrrr  rrrrrrrrrrrrrrrrrrrrrr\n"
-"\n"
-"       INSTRUCTION SETS WANT TO BE FREE\n";
+#include "platform_interface.h"
 
 void print_logo()
 {
-  putstring(logo);
+  putstring(platform__get_logo());
```
aswaterman: We probably should allow ```platform__get_logo``` to return NULL, so should check for that here.

palmer-dabbelt: Done.

PR #60, opened 2017-08-10
The DTB argument to boot_other_hart shouldn't be used in BBL

palmer-dabbelt:
I screwed up refactoring bbl.


palmer-dabbelt:
better?

PR #61, opened 2017-08-14
finisher: support terminating sifive devices simulation

terpstra:
Worked for me in simulating linux boot with a random call to die().


terpstra:
/scratch/terpstra/freedom-u-sdk/riscv-pk/machine/minit.c:143: Randomly stopping the program! 42
Power off

PR #67, opened 2017-11-02
Detect harts that can't boot Linux instead of hard-coding them

palmer-dabbelt:
This checks to see if a hart can't boot Linux by looking for a
compatible "mmu-type" field.  If the hart can't boot Linux, then bbl
masks it off.

### DIFF
--- a/bbl/bbl.c
+++ b/bbl/bbl.c
@@ -9,6 +9,7 @@
 #include <string.h>
 
 static const void* entry_point;
+long disabled_hart_mask;
 
 static uintptr_t dtb_output()
 {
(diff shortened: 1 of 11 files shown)


palmer-dabbelt:
@terpstra I just found this sitting on my computer somewhere, but it boots in spike... :)

terpstra:
We should change spike to put something indicating HTIF into DTS.

terpstra:
As for this PR, does it also boot the FPGAs? If so, let's merge it! 👍

palmer-dabbelt:
That's next on the list, but at least BBL tries the UART first.

palmer-dabbelt:
I haven't tried it on an FPGA...

palmer-dabbelt:
@aswaterman I just bumbled my way through https://github.com/riscv/riscv-pk/pull/67/commits/27498013ad7684a5f22a99790ac29badea205a3f without checking to see if this still results in sane FP emulation code.  Does it look OK?

palmer-dabbelt:
This boots on our U54-MC FPGA.  I can't get a DHCP lease, but it doesn't seem like it's related to this...

aswaterman:
It looks OK but you should try it out.  If you compile bbl with -march=rv64imac, and use spike --isa=rv64imac, then a floating-point-enabled Linux userland should still work correctly.

PR #68, opened 2017-11-03
Remove the platform interface

palmer-dabbelt:
We now automatically detect everything that the platform interface used
to be used for, so it's now obsolete!


machine/htif.c
```diff
@@ -85,3 +95,44 @@ void htif_poweroff()
...
+  struct htif_scan scan;
+
+  memset(&cb, 0, sizeof(cb));
+  cb.open = htif_open;
+  cb.prop = htif_prop;
+  cb.done = htif_done;
+  cb.extra = &scan;
+
+  fdt_scan(fdt, &cb);
+}
```
terpstra: It makes me a bit sad how much code it takes to extract a property from DTS. Without using dynamic memory, tho, I'm not sure how to improve things.

PR #72, opened 2017-12-13
Add a 16550 UART driver to back the SBI console

palmer-dabbelt:
QEMU currently provides the console via HTIF and the SBI.  That's a bit
messy because BBL polls for serial input, which means that typing too
quickly loses characters.  While QEMU has a standard 16550 device model,
there's no way to have two consoles share the console in QEMU (as they'd
step all over each other) so that means we can't have both the HTIF
console and the 16550 console.

With this patch, QEMU can be changed to use a 16650 instead of the HTIF
for serial output.  Linux will use the SBI for early printk support
(which is fine, polling for output is stable) and then swap over as soon
as it detects the UART.  When Linux swaps it prints out the whole
history, but there's probably a way to get around that.  There's a few
lines that are output to both, but it appears the Linux driver is close
enough to ours that nothing catastrophic happens -- there's not much to
the device, so hopefully that pans out on real hardware too.

Once Linux swaps over to natively using the driver we get reliable
console input.  If you don't have the in-kernel driver then Linux never
swaps over and keeps using the SBI console just like before.

### DIFF
--- a/machine/machine.mk.in
+++ b/machine/machine.mk.in
@@ -12,6 +12,7 @@ machine_hdrs = \
   mcall.h \
   mtrap.h \
   uart.h \
+  uart16550.h \
   finisher.h \
   unprivileged_memory.h \
   vm.h \
@@ -26,6 +27,7 @@ machine_c_srcs = \
   fp_emulation.c \
   fp_ldst.c \
   uart.c \
+  uart16550.c \
   finisher.c \
   misaligned_ldst.c \
 
--- a/machine/minit.c
+++ b/machine/minit.c
@@ -4,6 +4,7 @@
 #include "fp_emulation.h"
 #include "fdt.h"
 #include "uart.h"
+#include "uart16550.h"
 #include "finisher.h"
 #include "disabled_hart_mask.h"
 #include "htif.h"
@@ -141,6 +142,7 @@ void init_first_hart(uintptr_t hartid, uintptr_t dtb)
 {
   // Confirm console as early as possible
   query_uart(dtb);
+  query_uart16550(dtb);
   query_htif(dtb);
 
   hart_init();
--- a/machine/mtrap.c
+++ b/machine/mtrap.c
@@ -5,6 +5,7 @@
 #include "bits.h"
 #include "vm.h"
 #include "uart.h"
+#include "uart16550.h"
 #include "finisher.h"
 #include "fdt.h"
 #include "unprivileged_memory.h"
@@ -22,6 +23,8 @@ static uintptr_t mcall_console_putchar(uint8_t ch)
 {
   if (uart) {
     uart_putchar(ch);
+  } else if (uart16550) {
+    uart16550_putchar(ch);
   } else if (htif) {
     htif_console_putchar(ch);
   }
@@ -73,6 +76,8 @@ static uintptr_t mcall_console_getchar()
 {
   if (uart) {
     return uart_getchar();
+  } else if (uart16550) {
+    return uart16550_getchar();
   } else if (htif) {
     return htif_console_getchar();
   } else {
--- a/machine/uart16550.c
+++ b/machine/uart16550.c
@@ -1,0 +1,74 @@
+#include <string.h>
+#include "uart16550.h"
+#include "fdt.h"
+
+volatile uint8_t* uart16550;
+
+#define UART_REG_QUEUE     0
+#define UART_REG_LINESTAT  5
+#define UART_REG_STATUS_RX 0x01
+#define UART_REG_STATUS_TX 0x20
+
+void uart16550_putchar(uint8_t ch)
+{
+  while ((uart16550[UART_REG_LINESTAT] & UART_REG_STATUS_TX) == 0);
+  uart16550[UART_REG_QUEUE] = ch;
+}
+
+int uart16550_getchar()
+{
+  if (uart16550[UART_REG_LINESTAT] & UART_REG_STATUS_RX)
+    return uart16550[UART_REG_QUEUE];
+  return -1;
+}
+
+struct uart16550_scan
+{
+  int compat;
+  uint64_t reg;
+};
+
+static void uart16550_open(const struct fdt_scan_node *node, void *extra)
+{
+  struct uart16550_scan *scan = (struct uart16550_scan *)extra;
+  memset(scan, 0, sizeof(*scan));
+}
+
+static void uart16550_prop(const struct fdt_scan_prop *prop, void *extra)
+{
+  struct uart16550_scan *scan = (struct uart16550_scan *)extra;
+  if (!strcmp(prop->name, "compatible") && !strcmp((const char*)prop->value, "ns16550a")) {
+    scan->compat = 1;
+  } else if (!strcmp(prop->name, "reg")) {
+    fdt_get_address(prop->node->parent, prop->value, &scan->reg);
+  }
+}
+
+static void uart16550_done(const struct fdt_scan_node *node, void *extra)
+{
+  struct uart16550_scan *scan = (struct uart16550_scan *)extra;
+  if (!scan->compat || !scan->reg || uart16550) return;
+
+  uart16550 = (void*)(uintptr_t)scan->reg;
+  // http://wiki.osdev.org/Serial_Ports
+  uart16550[1] = 0x00;    // Disable all interrupts
+  uart16550[3] = 0x80;    // Enable DLAB (set baud rate divisor)
+  uart16550[0] = 0x03;    // Set divisor to 3 (lo byte) 38400 baud
+  uart16550[1] = 0x00;    //                  (hi byte)
+  uart16550[3] = 0x03;    // 8 bits, no parity, one stop bit
+  uart16550[2] = 0xC7;    // Enable FIFO, clear them, with 14-byte threshold
+}
+
+void query_uart16550(uintptr_t fdt)
+{
+  struct fdt_cb cb;
+  struct uart16550_scan scan;
+
+  memset(&cb, 0, sizeof(cb));
+  cb.open = uart16550_open;
+  cb.prop = uart16550_prop;
+  cb.done = uart16550_done;
+  cb.extra = &scan;
+
+  fdt_scan(fdt, &cb);
+}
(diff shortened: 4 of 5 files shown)


palmer-dabbelt:
@michaeljclark

michaeljclark:
I've made the associated change to the riscv-qemu virt board to make 16550a the default.

https://github.com/riscv/riscv-qemu/pull/82

This fixes copy/paste and up/down arrows in the QEMU console.

michaeljclark:
BTW I think we can and should fix HTIF for the QEMU spike boards at some point, perhaps not as the highest priority. We should probably get SiFiveUART working well for E300/U500 ahead of HTIF.

I made a temporary fix to HTIF in riscv-qemu to get console output working with recent bbl which broke due to a change in the HTIF acknowledgment protocol. The following commented out code was deadlocking due to a race, however, this is for console output, not input:

https://github.com/riscv/riscv-qemu/blob/5d93ae8f6408315197a8df07ca8a7177d00e1d43/hw/riscv/riscv_htif.c#L245-L258

We need a queue because as characters are sent, an interrupt is raised, but if another character is sent before the CPU wakes up, it overwrites the last character, so the guest only reads the last character. It's documented in the code as a TODO:

https://github.com/riscv/riscv-qemu/blob/5d93ae8f6408315197a8df07ca8a7177d00e1d43/hw/riscv/riscv_htif.c#L195

aswaterman:
Unfortunately I don’t have time to review this in detail, but it seems ok in concept.

michaeljclark:
I'll do some testing on this tomorrow... I can test in qemu with the spike board (HTIF) and the virt board (with UART).

It hopefully means we can have a kernel configured with CONFIG_HVC_RISCV_SBI, CONFIG_SERIAL_8250, CONFIG_SERIAL_8250_CONSOLE, CONFIG_SERIAL_OF_PLATFORM, etc and depending on which environment it's booted in, it should do the right thing...

michaeljclark:
The spike HTIF board has this in device-tree:

```
    htif {
        compatible = "ucb,htif0";
    };
```

and the virt board with 16550a UART has this in device-tree:

```
    chosen {
        stdout-path = "/uart@10000000";
    };
    uart@10000000 {
        interrupts = <0x0000000a>;
        interrupt-parent = <0x00000002>;
        clock-frequency = <0x00384000>;
        reg = <0x00000000 0x10000000 0x00000000 0x00000100>;
        compatible = "ns16550a";
    };
```

and the sifive_u500 board with SiFiveUART has this (which we still need to test):

```
    uart@10013000 {
        interrupts = <0x00000001>;
        interrupt-parent = <0x00000002>;
        reg = <0x00000000 0x10013000 0x00000000 0x00001000>;
        compatible = "sifive,uart0";
    };
```

We should also add the "chosen" node to the sifive_u500 board's device-tree as this is standard machinery in `linux/drivers/of/fdt.c` for selecting early console.

We should be able to test all three. Oh, we need a `linux/drivers/tty/serial/sifive_uart.c` for testing number 3?

palmer-dabbelt:
I'll add a Linux driver for the SiFive UART, but that's orthogonal to this PR.  BBL doesn't respect "chosen", but we could add that as a feature.

michaeljclark:
I've tested the patch with the qemu spike_v1.10 board and the virt board. It seems okay to me.

Spike board using HTIF:

```
$ qemu-system-riscv64 -nographic -machine spike_v1.10 -kernel bbl-1.10-hvc
              vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv
                  vvvvvvvvvvvvvvvvvvvvvvvvvvvv
rrrrrrrrrrrrr       vvvvvvvvvvvvvvvvvvvvvvvvvv
rrrrrrrrrrrrrrrr      vvvvvvvvvvvvvvvvvvvvvvvv
rrrrrrrrrrrrrrrrrr    vvvvvvvvvvvvvvvvvvvvvvvv
rrrrrrrrrrrrrrrrrr    vvvvvvvvvvvvvvvvvvvvvvvv
rrrrrrrrrrrrrrrrrr    vvvvvvvvvvvvvvvvvvvvvvvv
rrrrrrrrrrrrrrrr      vvvvvvvvvvvvvvvvvvvvvv  
rrrrrrrrrrrrr       vvvvvvvvvvvvvvvvvvvvvv    
rr                vvvvvvvvvvvvvvvvvvvvvv      
rr            vvvvvvvvvvvvvvvvvvvvvvvv      rr
rrrr      vvvvvvvvvvvvvvvvvvvvvvvvvv      rrrr
rrrrrr      vvvvvvvvvvvvvvvvvvvvvv      rrrrrr
rrrrrrrr      vvvvvvvvvvvvvvvvvv      rrrrrrrr
rrrrrrrrrr      vvvvvvvvvvvvvv      rrrrrrrrrr
rrrrrrrrrrrr      vvvvvvvvvv      rrrrrrrrrrrr
rrrrrrrrrrrrrr      vvvvvv      rrrrrrrrrrrrrr
rrrrrrrrrrrrrrrr      vv      rrrrrrrrrrrrrrrr
rrrrrrrrrrrrrrrrrr          rrrrrrrrrrrrrrrrrr
rrrrrrrrrrrrrrrrrrrr      rrrrrrrrrrrrrrrrrrrr
rrrrrrrrrrrrrrrrrrrrrr  rrrrrrrrrrrrrrrrrrrrrr

       INSTRUCTION SETS WANT TO BE FREE
[    0.000000] OF: fdt: Ignoring memory range 0x80000000 - 0x80200000
[    0.000000] Linux version 4.15.0-rc3-01064-g418457520fbb (mclark@minty) (gcc version 7.1.1 20170509 (GCC)) #51 Thu Dec 14 09:20:58 NZDT 2017
[    0.000000] bootconsole [early0] enabled
[    0.000000] Initial ramdisk at: 0x        (ptrval) (699370 bytes)
[    0.000000] Zone ranges:
[    0.000000]   DMA      [mem 0x0000000080200000-0x0000000087ffffff]
[    0.000000]   Normal   empty
[    0.000000] Movable zone start for each node
[    0.000000] Early memory node ranges
[    0.000000]   node   0: [mem 0x0000000080200000-0x0000000087ffffff]
[    0.000000] Initmem setup node 0 [mem 0x0000000080200000-0x0000000087ffffff]
[    0.000000] elf_hwcap is 0x112d
[    0.000000] Built 1 zonelists, mobility grouping on.  Total pages: 31815
```

QEMU Virtio Board using 16550:

```
$ qemu-system-riscv64 -nographic -machine virt -kernel bbl-1.10-uart 
              vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv
                  vvvvvvvvvvvvvvvvvvvvvvvvvvvv
rrrrrrrrrrrrr       vvvvvvvvvvvvvvvvvvvvvvvvvv
rrrrrrrrrrrrrrrr      vvvvvvvvvvvvvvvvvvvvvvvv
rrrrrrrrrrrrrrrrrr    vvvvvvvvvvvvvvvvvvvvvvvv
rrrrrrrrrrrrrrrrrr    vvvvvvvvvvvvvvvvvvvvvvvv
rrrrrrrrrrrrrrrrrr    vvvvvvvvvvvvvvvvvvvvvvvv
rrrrrrrrrrrrrrrr      vvvvvvvvvvvvvvvvvvvvvv  
rrrrrrrrrrrrr       vvvvvvvvvvvvvvvvvvvvvv    
rr                vvvvvvvvvvvvvvvvvvvvvv      
rr            vvvvvvvvvvvvvvvvvvvvvvvv      rr
rrrr      vvvvvvvvvvvvvvvvvvvvvvvvvv      rrrr
rrrrrr      vvvvvvvvvvvvvvvvvvvvvv      rrrrrr
rrrrrrrr      vvvvvvvvvvvvvvvvvv      rrrrrrrr
rrrrrrrrrr      vvvvvvvvvvvvvv      rrrrrrrrrr
rrrrrrrrrrrr      vvvvvvvvvv      rrrrrrrrrrrr
rrrrrrrrrrrrrr      vvvvvv      rrrrrrrrrrrrrr
rrrrrrrrrrrrrrrr      vv      rrrrrrrrrrrrrrrr
rrrrrrrrrrrrrrrrrr          rrrrrrrrrrrrrrrrrr
rrrrrrrrrrrrrrrrrrrr      rrrrrrrrrrrrrrrrrrrr
rrrrrrrrrrrrrrrrrrrrrr  rrrrrrrrrrrrrrrrrrrrrr

       INSTRUCTION SETS WANT TO BE FREE
[    0.000000] OF: fdt: Ignoring memory range 0x80000000 - 0x80200000
[    0.000000] Linux version 4.15.0-rc3-01064-g418457520fbb (mclark@minty) (gcc version 7.1.1 20170509 (GCC)) #50 Thu Dec 14 09:18:51 NZDT 2017
[    0.000000] bootconsole [early0] enabled
[    0.000000] Initial ramdisk at: 0x        (ptrval) (699370 bytes)
[    0.000000] Zone ranges:
[    0.000000]   DMA      [mem 0x0000000080200000-0x0000000087ffffff]
[    0.000000]   Normal   empty
[    0.000000] Movable zone start for each node
[    0.000000] Early memory node ranges
[    0.000000]   node   0: [mem 0x0000000080200000-0x0000000087ffffff]
[    0.000000] Initmem setup node 0 [mem 0x0000000080200000-0x0000000087ffffff]
[    0.000000] elf_hwcap is 0x112d
[    0.000000] Built 1 zonelists, mobility grouping on.  Total pages: 31815
```

palmer-dabbelt:
You should be able to boot the same BBL and Linux images on both boards: if Linux doesn't see a 16550 in the device tree then it'll just keep using the SBI terminal, which BBL will provide by either HTIF or the SiFive UART depending on which is emulated.

michaeljclark:
I was able to use the same kernel config but I had an issue with the spike board. I needed to add console=hvc0 on the kernel command line. Other than that, it was exactly the same kernel config between both boards.

michaeljclark:
with the Virt board it's possible to use either SBI or uart for early console.

palmer-dabbelt:
OK, well, it sounds like this might be either a Linux or QEMU device tree problem.  I think it's not a BBL problem, so I'm going to merge this.

michaeljclark:
OK. Good. With this commit, SBI console works fine with either HTIF or 16550a...

PR #84, opened 2018-02-20
32-bit HTIF Fix: TOHOST_CMD command word is 64-bit on both rv32 and rv64

heshamelmatary:
That's a pending fix that I had to do to be able to run 32-bit versions of RTEMS and seL4. 

Not quite sure that's the right fix to do though

### DIFF
--- a/machine/htif.h
+++ b/machine/htif.h
@@ -3,14 +3,9 @@
 
 #include <stdint.h>
 
-#if __riscv_xlen == 64
 # define TOHOST_CMD(dev, cmd, payload) \
   (((uint64_t)(dev) << 56) | ((uint64_t)(cmd) << 48) | (uint64_t)(payload))
-#else
-# define TOHOST_CMD(dev, cmd, payload) ({ \
-  if ((dev) || (cmd)) __builtin_trap(); \
-  (payload); })
-#endif
+
 #define FROMHOST_DEV(fromhost_value) ((uint64_t)(fromhost_value) >> 56)
 #define FROMHOST_CMD(fromhost_value) ((uint64_t)(fromhost_value) << 8 >> 56)
 #define FROMHOST_DATA(fromhost_value) ((uint64_t)(fromhost_value) << 16 >> 16)


heshamelmatary:
I see. I guess the same situation holds on 32-bit Linux version when/if run on Spike?

sorear:
@aswaterman Isn't that what `htif_lock` is for?

aswaterman:
That’s for synchronization between target threads. This is a race condition with the host machine. It’s not unsolvable, just needs some thought (and possible reworking of fesvr code).

palmer-dabbelt:
We might be able to construct something that's race free here, but I don't think it's worth bending over backwards for HTIF now that we have real platforms.  I think you'll be better off just using QEMU and targeting real systems as those won't need HTIF at all.

I'm also not sure why this code fixes anything: it'll either trap (and presumably break) or produce the same result as the original code (which has the race condition that Andrew mentioned).

What is the actual error behavior you're seeing?

aswaterman:
It's the other way around: the old code traps, rather than exhibiting the race; this patch removes the trap, so it appears to work, but it is racy.

michaeljclark:
I kind of disagree with disabling a feature just because its function depends on undefined behavior. In fact, the qemu HTIF driver receives tohost/fromhost reads/writes in 32-bit chunks. Essentially the undefined behavior we depend on is ordered writes, with the low order word written first. That is how the QEMU driver is written to handle 32-bit HTIF. It works on x86_64 TSO. We essentially need to split the writes and use 32-bit atomic_store with the low word first. It's not racy if the order is defined if one reasons about the code. i.e. cmd and dev is only read after the payload is written.

I don't mind to come up with a complete patch that works, not by depending on implementation-defined behaviour, rather uses explicit ordering of the reads and writes...

michaeljclark:
Sure it needs some more work, but it minimum it will let us test riscv32 linux in spike and the qemu spike machine, as "it works" but is racy. It's racy in a defined way on a TSO machine where the compiler emits the store of the low word first. Yes, we are depending on implementation-defined behaviour for it to work, but that may be sufficient to allow testing (without having to patch bbl for the tests).

michaeljclark:
If I get time I might look into solving this, in such a way as preserving backwards compatibility for HTIF on RV64. A two register message passing interface between otherwise paritioned processors (no common address space) is a useful mechanism. Apple uses a similar technique to communicator between the application processors running iOS and seL4 on their secure enclave processor. I think we should keep HTIF and make it a bit more robust and document it... when we get time...

Currently the acknowledgement mechanism is broken in QEMU so its even racy on RV64. i.e. drops characters if the host doesn't read them before another character is written. The HTIF in QEMU was hacked due to a change in behaviour to spike circa priv 1.9.1 and spike present. HTIF in QEMU works to the extent that we can boot a kernel and console input works (somewhat).

I do think we should fix it... eventually...

michaeljclark:
You might not have seen this:

https://github.com/riscv/riscv-qemu/blob/qemu-upstream-v6/hw/riscv/riscv_htif.c#L178-L223

QEMU assumes low order word is written first, which is apparantly what is happening.

The acknowledgement however is currently broken so its even racy on riscv64.

sorear:
I don't have strong feelings about this patch or what we do for HTIF, but I do want to see an agreement reached here so that we can get official support for riscv32 soon.

PR #85, opened 2018-02-25
Allow 32-bit HTIF because it works on qemu-system-riscv32

michaeljclark:
To enable testing of rv32 riscv-linux on qemu-system-riscv32
it would be good if we had working HTIF in bbl. This change
has been tested using riscv-probe in spike and qemu:

- https://github.com/michaeljclark/riscv-probe

While there are compiler issues that could re-order 64-bit
loads and stores, and there is no official HTIF over 32-bit
protocol, or 64-bit HTIF with ordered loads/stores and fences;
however, given that it works as it is, it should be safe to
re-enable 32-bit support with a caveat.

Signed-off: mjc@sifive.com

### DIFF
--- a/machine/htif.h
+++ b/machine/htif.h
@@ -3,14 +3,9 @@
 
 #include <stdint.h>
 
-#if __riscv_xlen == 64
-# define TOHOST_CMD(dev, cmd, payload) \
+/* Caveat: HTIF on rv32 is not officially supported. Use at your own risk */
+#define TOHOST_CMD(dev, cmd, payload) \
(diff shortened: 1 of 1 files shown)


michaeljclark:
This is so I can test riscv-linux on rv32...

aswaterman:
This is the same as #84.  I'm reluctant to merge this unless we eliminate the race condition.  What do you thin, @palmer-dabbelt?

michaeljclark:
I didn't look at the PRs first. PR https://github.com/riscv/riscv-pk/pull/84 is the same so I'm closing this PR.

PR #89, opened 2018-03-16
Changed install_subdir to $host_alias with substituted 64->32 instead…

rishikhan:
Changed install_subdir to $host_alias with substituted 64->32 instead of hardcoded riscv[32|64]-unknown-elf

This is to allow other compilers like riscv64-VENDOR-OS ... and not just unknown-elf

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,7 @@ By default, 64-bit (RV64) versions of `pk` and `bbl` are built.  To
 built 32-bit (RV32) versions, supply a `--enable-32bit` flag to the
 configure command.
 
-The `install` step installs 64-bit build products into
-`$RISCV/riscv64-unknown-elf`, and 32-bit versions into
-`$RISCV/riscv32-unknown-elf`.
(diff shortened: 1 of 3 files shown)


rishikhan:
Well, the instructions say you can do --host=riscv64-unknown-elf --enable-32bit
and this will use -m32 to compile and put it in the riscv32-unknown-elf directory. I was trying to stay compatible with the original semantics. Let me know how you want to handle it? Keep the semantics, or update the README?

seldridge:
Got it. No, the README should be right, here. Following along with that... the README.md should be updated to indicate the new installation location.

rishikhan:
Let me know if you need anything else to merge. Thanks!

seldridge:
Will need an actual owner to approve, but it looks good to me.

aswaterman:
-m32 isn't the right way to do this anymore, so --enable-32bit is completely broken.  But that is a separate issue, so I'll merge this as-is.

PR #90, opened 2018-03-21
Fix build with compilers defaulting to PIE

aurel32:
Debian toolchain defaults to PIE, and I guess that will also be the case
of most distributions. This cause bbl to be non-functional.

This patch fixes that by adding -fno-PIE in the default CFLAGS.


aswaterman:
thanks

Issue #92, opened 2018-03-28
`spike pk hello` does not terminate (32bit)


Generalbrus:
On macOS, I have built and installed the 32bit version of pk and of the riscv-gnu-toolchain.
I manage to compile programs using `riscv32-unknown-elf-gcc`, but when running `spike -l path/to/pk32 hello`,  the script never stops, and if I halt it manually I get `exception trap_instruction_access_fault`.

I've looked for the problem online, and my impression is that the 32bit version of the toolchain is not really that stable. I haven't found a solution for my issue.
Am I doing something wrong?

gmarkall:
I'm seeing something similar with a 32-bit build - reverting this commit seems to fix the issue: https://github.com/riscv/riscv-pk/commit/5a0e3e55cab1d6f3462c77e852f16365c79f2c98

As far as I can tell from looking at the log generated by Spike, something appears to go wrong from that `printm` with pk ending up in `trap_from_machine_mode` after the `ebreak` instruction in `htif_console_putchar` - this then ends up in `bad_trap` which tries to print, which ends up in `htif_console_putchar` again which spins forever trying to acquire `htif_lock`.

@Generalbrus does commenting out the `printm` added in the above commit and rebuilding fix the issue for you?

gmarkall:
I hadn't realised when I wrote my earlier comment that the things going wrong were probably due to lack of support for HTIF (and therefore `printm`) on rv32 - PR #94 created to try and workaround the issue.

Generalbrus:
@gmarkall I tried applying the edits in both commits but I still get `trap_instruction_access_fault` and `trap_illegal_instruction`...

gmarkall:
I guess there's some other issue as well - does it get as far as printing "Hello" (assuming your hello program is intended to print "Hello")?

Generalbrus:
It loops indefinitely... The hello program is the one mentioned in the `riscv-tools` README to test everything is fine, it just prints "Hello World".

gmarkall:
@Generalbrus I just used `configure --prefix=<install path> --host=riscv32-unknown-elf` because I already had a toolchain with `rv32im` as the default arch.

G-ram:
I also have this issue with the latest and removing the printm does in fact solve this issue.

PR #93, opened 2018-03-29
Remove --enable-32bit in favour of --with-arch and --with-abi

gmarkall:
Noting that a previous PR to fix the `--enable-32bit` flag (https://github.com/riscv/riscv-pk/pull/70) does un-break `--enable-32bit` but always configures for specific ISA strings and ABIs (mentioned in https://github.com/riscv/riscv-gnu-toolchain/issues/294), this PR fixes building for 32-bit by removing `--enable-32bit` entirely and instead adding `--with-arch` and `--with-abi` with the "usual" meanings.

I've tried this with various combinations, some of which are:

* `riscv64-unknown-elf` as host and `--with-arch=rv32im` (forces 32-bit install location)
* `riscv64-unknown-elf` as host, no arch or abi flags
* `riscv32-unknown-elf` as host (uses 32-bit install location, but not forced)
* `riscv32-unknown-elf` as host, 32-bit arch and abi flags specified
* `riscv64-unknown-elf` as host, 64-bit arch and abi flasg specified
* etc..

All these combinations (and some others not listed) appeared to do the right thing. I have been testing with the commit https://github.com/riscv/riscv-pk/commit/5a0e3e55cab1d6f3462c77e852f16365c79f2c98 reverted, as it appears to break 32-bit: https://github.com/riscv/riscv-pk/issues/92 - however, I don't believe there's any relation between this PR and that commit. I'm not quite sure what the fix for #92 on rv32 is yet.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,18 +24,18 @@ directory is not supported; you need to use a separate build directory.
 
     $ mkdir build
     $ cd build
-    $ ../configure --prefix=$RISCV --host=riscv64-unknown-elf
+    $ ../configure --prefix=$RISCV --host=riscv64-unknown-elf \
+                   --with-arch=rv64imafd \
+                   --with-abi=lp64d
     $ make
     $ make install
 
-Alternatively, the GNU/Linux toolchain may be used to build this package,
-by setting `--host=riscv64-unknown-linux-gnu`.
-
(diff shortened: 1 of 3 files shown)


gmarkall:
Superseded by #102.

PR #94, opened 2018-03-29
Don't printm on rv32

gmarkall:
As far as I could understand, the recently-added `printm` seems to prevent `pk` being usable at all on rv32 (e.g. https://github.com/riscv/riscv-pk/issues/92) because HTIF is not supported. Although there is some discussion and movement towards supporting the HTIF on rv32 in https://github.com/riscv/riscv-pk/pull/84, it would be good if `pk` didn't call `printm` on rv32 in the meantime so that it is usable again for rv32. This commit makes that so - however, I would be happy to revise it if there are other suggestions for how this should be worked around.

Tested with riscv32-unknown-elf and riscv64-unknown-elf - the string "bbl loader" only appears on riscv64, and riscv32 programs seem to execute and terminate normally.

### DIFF
--- a/machine/minit.c
+++ b/machine/minit.c
@@ -143,7 +143,12 @@ void init_first_hart(uintptr_t hartid, uintptr_t dtb)
   query_uart(dtb);
   query_uart16550(dtb);
   query_htif(dtb);
+
+  // This is a workaround for a race condition. Presently printm does not work
+  // on rv32 because HTIF is not supported, so don't use the workaround on rv32.
+#if __riscv_xlen != 32
   printm("bbl loader\r\n");
+#endif
 
   hart_init();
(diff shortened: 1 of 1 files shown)


gmarkall:
@palmer-dabbelt Quick ping following our chat at the RISC-V workshop... I think this is a fairly trivial fix to get pk working on spike for rv32 again, but do let me know if that's not the case for some reason.

gmarkall:
From the discussion on #84, it looks like it's not clear how to make HTIF work on rv32, so I'm not sure how to make `printm` work. What do you think the definition of `printm` should be changed to?

aswaterman:
Sorry I let this fall by the wayside.  I'm opening a new PR that addresses the root cause.

PR #96, opened 2018-03-30
Fix for missing supervisor mode when running on E51

jstraus59:
The E51 core on the U54-MC lacks supervisor mode, thus the plic_s_ie and plic_s_thresh are NULL when running on this core. This adds checks for this case.

Running bbl on the OVP model of the U54-MC device resulted in a null pointer error in the hart_plic_init() function. 

This occured because bbl ran on hart0 which is the E51 core  (at least in this model), which does not implement a supervisor mode. Thus there is no entry in the PLIC node of the device tree that sets the pointers for supervisor mode, plic_s_ie and plic_s_thresh, in the hls_t structure.

This error would not occur if the bbl were running on one of the U54 cores, but the [device documentation](https://www.sifive.com/documentation/risc-v-core/u54-mc-risc-v-core-ip-manual/) indicates hart0 should be the E51 and the bbl code specifically selects the hart with mhartid==0 to run on.

### DIFF
--- a/machine/minit.c
+++ b/machine/minit.c
@@ -124,10 +124,17 @@ static void hart_plic_init()
     return;
 
   size_t ie_words = plic_ndevs / sizeof(uintptr_t) + 1;
-  for (size_t i = 0; i < ie_words; i++)
-    HLS()->plic_s_ie[i] = ULONG_MAX;
+  for (size_t i = 0; i < ie_words; i++) {
+     if (HLS()->plic_s_ie) {
+        // Supervisor not always present
+        HLS()->plic_s_ie[i] = ULONG_MAX;
+     }
+  }
   *HLS()->plic_m_thresh = 1;
-  *HLS()->plic_s_thresh = 0;
+  if (HLS()->plic_s_thresh) {
+      // Supervisor not always present
+      *HLS()->plic_s_thresh = 0;
+  }
 }
 
 static void wake_harts()


jstraus59:
I have a little more detail on this problem, including a theory on why it does not cause a problem in the hardware.

1) The E51 hart in the U54-MC lacks supervisor mode, so the plic does not include a region for supervisor mode registers for hart0.
2) Thus in the device tree the interrupts-extended array for the interrupt controller does not include an entry with the value IRQ_S_EXT (=9) for the hart0. This means that the plic_s_ir and plic_s_thresh fields in the hls_t struct for hart0 are never initialized by the function plic_done() (at line 438 in [machine/fdt.c](https://github.com/riscv/riscv-pk/blob/master/machine/fdt.c#L438)).
3) This results in a write to memory  addresses 0x00:0x07 by the function hart_plic_init() (at line 199 in [machine/minit.c](https://github.com/riscv/riscv-pk/blob/master/machine/minit.c#L128)).

Apparently writing to address 0 does not cause an exception in the HiFiveUnleashed board. Looking at the memory map I see that addresses 0x00:0xff are in the debug region and are described as Reserved. 

This problem is caught in the OVP FU540 virtual platform. I have worked around this problem by adding 256 bytes of memory at address 0, but I do think this is a real bug that should be fixed with this patch.

jstraus59:
One more follow up on this - I just spotted section 23.3.4 in the [FU540 Manual](https://www.sifive.com/documentation/chips/freedom-u540-c000-manual/) that states that address 0x0 is a "Safe Zero Address" where writes have no impact and reads return 0. This explains why the bbl works as-is on the HiFiveUnleashed board. 

However I think this patch needs to be applied so that the bbl will not fail on systems with a processor that lacks S mode and that cause an exception when address 0x0 is written.

PR #97, opened 2018-04-04
SBI calls for perf counters

solomatnikov:
compilable, no functional testing yet


aswaterman:
Unless I'm missing something, I don't think SBI_READ_PERF_CNT is necessary.  The OS should just read the hpmcounterX CSR directly.  The OS knows which counters are legal to read directly, because it used SBI_WRITE_PERF_CFG to program them, and that function returns 0 for valid counters.

@palmer-dabbelt I assume it's OK for context-switching a perf-counting process to be a slow path, so doing a few calls to SBI_WRITE_PERF_CFG is OK?

solomatnikov:
I was under impression that the kernel is not supposed to read perf counters directly b/c it would create virtualization hole.

palmer-dabbelt:
In this case maybe we don't need an SBI interface at all: just let the supervisor touch the CSRs directly and then have hypervisor mode swap them out if they're in use.  This would make the common cases fast -- it'd essentially be like lazy floating-point swapping, just for the performance counters.

solomatnikov:
If I understand it correctly, mcounteren[x] only allows reads of counter values, not writes to the config, so SBI_WRITE_PERF_CFG is still needed.

@aswaterman are mcounteren[x] bits set to 1 at reset? Or is additional init code required in SBI?

BTW, @richardxia asked about reset values of perf counters and perf control CSRs

aswaterman:
Something like SBI_WRITE_PERF_CFG is still necessary to set the counter configuration.

mcounteren is not reset by HW, but pk/bbl initializes it to all 1s (in machine/minit.c).

Per @richardxia's question, hpmcounters are not reset, but hpmevents are reset to 0.

I agree with @palmer-dabbelt that lazy swapping is possible without any SBI calls to read the counter values themselves.

solomatnikov:
Kernel perf https://lkml.org/lkml/2018/4/2/174 already only does read of perf counters on context switch, so if  kernel can read perf counters w/o SBI, then no SBI calls needed for context switch.

What about reset values for cycle and instret?

aswaterman:
Rocket and SiFive implementations all reset cycle and instret, but the ISA does not actually mandate this.

solomatnikov:
So SBI_WRITE_PERF_CNT would be needed to reset perf counter to a known value and prevent immediate overflow.

I presume cycle and instret are writable, so they can be written by CSR write as well.

aswaterman:
cycle and instret are not writable.  M-mode can write them via the mcycle and minstret CSRs.  So these should also be written via SBI_WRITE_PERF_CNT.

Which gets me to the other thought: we should keep the arguments to these SBI calls consistent with the CSR numbers, so id=0 means cycle, id=1 means instret=1, id=3 means hpmcounter3=3, etc.

solomatnikov:
> Which gets me to the other thought: we should keep the arguments to these SBI calls consistent with the CSR numbers, so id=0 means cycle, id=1 means instret=1, id=3 means hpmcounter3=3, etc.

yes, this makes sense

solomatnikov:
If CSR reads can be handled by traps into SBI, then CSR writes can also be handled by traps, right?

So, does it even makes sense to do SBI calls for this?

aswaterman:
Yeah, no new SBI calls seems like a cleaner option to me!

Since BBL trusts the OS, it can unconditionally perform the write on behalf of the OS.  We already have infrastructure for this - see the emulate_read_csr and emulate_write_csr functions.

A real hypervisor would emulate the write only if it thought the value was OK; if not, it would send the illegal-instruction trap to the OS to indicate a bad value.

solomatnikov:
I don't understand emulate_read_csr() code: for time CSR, it does not do read_csr(mtime) - is this code only used when time CSR does not exist in HW?

Also, there is no emulation code for cycle CSR.

PR #98, opened 2018-04-06
SBI emulation of reads and writes to perf counters and config

solomatnikov:


palmer-dabbelt:
Why are only HPM counter 3 and 4 emulated?

solomatnikov:
The chip has 2 programmable perf counters: 3 and 4. I guess cycle is 0 and instret is 1.

PR #99, opened 2018-04-10
Fix the wrong access of ipi pending address

zongbox:
The MENTRY_IPI_PENDING_OFFSET offset is based on stack pointer.


aswaterman:
Good catch, thanks.

PR #101, opened 2018-04-30
Replace compile option m32 by march/mabi

zongbox:
Use gcc -v to get the march and mabi option from toolchain.

The --with-arch is exist certainly in toolchain,
but the --with-abi is uncertain.

If the --with-abi is not exist in toolchain,
using 'g' and 'd' extension in arch to deduce the abi

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -52,12 +52,17 @@ endif
 
 prefix       := @prefix@
 enable_stow  := @enable_stow@
+ifeq (@BUILD_32BIT@, yes)
+  install_subdir := $(shell echo @host_alias@ | sed -e 's/64/32/g')
+else
+  install_subdir := @host_alias@
+endif
 
 ifeq ($(enable_stow),yes)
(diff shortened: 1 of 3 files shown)


aswaterman:
Upon some more thought, I think we should get rid of the --enable-32bit option and switch to using --with-arch, which is more standard because it matches our GCC build.  So you'd do something like --with-arch=rv32imac.  If --with-arch is not specified, it defaults to whatever the compiler's default is.

Now is a good time to make this change because it won't break backwards compatibility, since --enable-32bit has been broken for so long.  What do you think @palmer-dabbelt ?

I believe --with-abi is not necessary for this project.  It should be OK to unconditionally compile it with a no-float ABI, i.e., ilp32 for 32-bit and lp64 for 64-bit.

palmer-dabbelt:
That makes sense to me -- since `--enable-32bit` doesn't work then it's not really an interface break :).

zongbox:
I re-send the patch for new mechanism as discussion

PR #102, opened 2018-05-01
Replace the --enable-32bit option by --with-arch

zongbox:
Get rid of the --enable-32bit option and switch to use --with-arch,
which is more standard because it matches the GCC build.

If --with-arch is not specified, it defaults to whatever the
compiler's default is.

The --with-abi is not necessary for this project.
Unconditionally compile it with a no-float ABI.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -48,6 +48,19 @@ else
   project_ver:=@PACKAGE_VERSION@
 endif
 
+# If --with-arch is not specified, it defaults to whatever the compiler's
+# default is. The -with-abi is not necessary for this project. Unconditionally
+# compile it with a no-float ABI. i.e., ilp32 for 32-bit and lp64 for 64-bit.
+
+ifeq (@WITH_ARCH@,)
+  march := -march=$(shell @CC@ -v 2>&1 | \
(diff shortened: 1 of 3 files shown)


gmarkall:
@palmer-dabbelt Pinging you about this PR following our chat at the RISC-V workshop - I did have #93 open to do something similar, but this probably makes more sense as #93 included the un-needed `--with-abi` flag.

zongbox:
Just rebase to latest commit

andreas-schwab:
riscv64-linux-gcc: error: missing argument to ‘-march=’

andreas-schwab:
/daten/cross/riscv64-linux/riscv64-linux/sys-root/usr/include/gnu/stubs.h:8:11: fatal error: gnu/stubs-lp64.h: No such file or directory
 # include <gnu/stubs-lp64.h>
           ^~~~~~~~~~~~~~~~~~
compilation terminated.
make: *** [Makefile:320: fdt.o] Error 1

zongbox:
As discussion before, we compile unconditionally it with a no-float ABI.
@aswaterman whether we need to deduce the abi like last PR to support float ABI?

PR #104, opened 2018-05-14
bbl: boot payload in machine mode on --enable-boot-machine

heshamelmatary:
If --disable-vm is passed, BBL disables VM and runs the payload in machine mode.
This is useful for payloads (e.g. RTOSes or other OSes) that want to run
only in machine mode while still relying on bbl/pk for system calls and emulation

When passed, --disable-vm implicitly runs the payload in machine mode

### DIFF
--- a/machine/minit.c
+++ b/machine/minit.c
@@ -95,7 +95,9 @@ static void hart_init()
 {
   mstatus_init();
   fp_init();
+#ifndef BBL_BOOT_MACHINE
   delegate_traps();
+#endif /* BBL_BOOT_MACHINE */
 }
 
 static void plic_init()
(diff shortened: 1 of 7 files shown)


aswaterman:
Seems like a form of chain loading, which is potentially useful, but I'd like to hear what Palmer thinks

palmer-dabbelt:
I didn't look at the code, but the idea seems good.

palmer-dabbelt:
Do you mind resolving the `configure` conflict?  Sorry this took a while to look at.

machine/minit.c
```diff
@@ -197,3 +199,27 @@ void enter_supervisor_mode(void (*fn)(uintptr_t), uintptr_t arg0, uintptr_t arg1
...
+
+void enter_machine_mode(void (*fn)(uintptr_t, uintptr_t), uintptr_t arg0, uintptr_t arg1)
+{
+  // Set up a PMP to permit access to all of memory.
+  // Ignore the illegal-instruction trap if PMPs aren't supported.
+  uintptr_t pmpc = PMP_NAPOT | PMP_R | PMP_W | PMP_X;
+  asm volatile ("la t0, 1f\n\t"
+                "csrrw t0, mtvec, t0\n\t"
+                "csrw pmpaddr0, %1\n\t"
+                "csrw pmpcfg0, %0\n\t"
```
palmer-dabbelt: Can you split this out into something like `setup_pmps(...)` so we don't duplicate this?

heshamelmatary:
@palmer-dabbelt done

PR #106, opened 2018-05-19
machine,minit: initialize emulated FCSR in enter_supervisor_mode()

sashimi-yzh:
* If BBL emulates the FPU, the trap handler will load emulated FCSR from
  x0's save slot into tp. The emulated FCSR should be initialized, else
  the field of rounding mode will contain garbage codes. This will
  lead to raising SIGABRT for a user mode program which tries to print a
  floating point variable. In glibc, __printf_fp_l() (defined in
  riscv-glibc/stdio-common/printf_fp.c) will call round_away() (defined
  in riscv-glibc/include/rounding-mode.h). With a garbage rounding mode
  in emulated FCSR, round_away() may call abort().

### DIFF
--- a/machine/minit.c
+++ b/machine/minit.c
@@ -197,6 +197,10 @@ void enter_supervisor_mode(void (*fn)(uintptr_t), uintptr_t arg0, uintptr_t arg1
   mstatus = INSERT_FIELD(mstatus, MSTATUS_MPIE, 0);
   write_csr(mstatus, mstatus);
   write_csr(mscratch, MACHINE_STACK_TOP() - MENTRY_FRAME_SIZE);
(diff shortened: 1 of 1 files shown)


sashimi-yzh:
I found this bug in the priv-1.9.1 version of bbl. But I have not tested it in the master version, due to the huge difference of everything between the master version and the version I use. Please confirm it.

aswaterman:
Makes sense, but I haven’t tested it yet.

aswaterman:
@palmer-dabbelt related issue: the Unix ABI initializes the fcsr. Can we verify our assumptions and document this in the ABI?

palmer-dabbelt:
I'm OK with that.

PR #108, opened 2018-05-23
Support timer interrupt pending clear in SBI

atishp04:
As per the privileged spec, timer interrupt pending bit should
be cleared by calling into SEE. Currently, SBI doesn't support
a method to achieve that.

Introduce a method to clear the timer interrupt pending bit in
interrupt pending csr.

Signed-off-by: Atish Patra <atish.patra@wdc.com>

### DIFF
--- a/machine/mcall.h
+++ b/machine/mcall.h
@@ -10,5 +10,6 @@
 #define SBI_REMOTE_SFENCE_VMA 6
 #define SBI_REMOTE_SFENCE_VMA_ASID 7
 #define SBI_SHUTDOWN 8
+#define SBI_CLEAR_TIP 9
 
 #endif
--- a/machine/mtrap.c
+++ b/machine/mtrap.c
@@ -79,6 +79,11 @@ static uintptr_t mcall_clear_ipi()
   return clear_csr(mip, MIP_SSIP) & MIP_SSIP;
 }
 
+static uintptr_t mcall_clear_tip()
+{
+  return clear_csr(mip, MIP_STIP);
+}
+
 static uintptr_t mcall_shutdown()
 {
   poweroff(0);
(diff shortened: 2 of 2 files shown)


aswaterman:
sbi_set_timer is the intended way to do this.  I'm not sure why you'd want to clear the current timer interrupt without scheduling another one.  If you did, though, there are two ways to do this:

- Simply disable timer interrupts by clearing `sie.STIE`.  Then `sip.STIP` is a don't-care.
- Schedule a timer interrupt infinitely far in the future (`sbi_set_timer((uint64_t)-1)`).

atishp04:
Clearing the timer interrupt bit in sbi_set_timer creates issues in nohz mode. As we are in nohz mode, the timer is not reprogrammed, the pending bits are not cleared. Thus timer interrupt kept on firing resulting cpu stalls.

Here are more details on the cpu stalls.
https://github.com/riscv/riscv-linux/issues/127

aswaterman:
I'm still not clear on why disabling the timer, or scheduling a timer interrupt infinitely far in the future, isn't sufficient.

palmer-dabbelt:
I agree with Andrew.

atishp04:
I am trying to think how to do that without introducing a hack in kernel. Because nohz mode should work seamlessly without any arch specific things.
I couldn't find any genuine arch hookup to do these two.

tick_nohz_handler() just returns without setting the timer again.
https://github.com/riscv/riscv-linux/blob/master/kernel/time/tick-sched.c#L1176

We can do checks in the timer interrupt handler to verify if nohz mode is set but I believe that may conflict with timer management.

Any suggestions?

atishp04:
And just for my knowledge, Why it is not recommended to clear the timer interrupt pending bit?
I couldn't find anything specific in the spec except this line.
"An SBI call to the SEE may be used to clear the pending timer interrupt."

That's what lead to me believe that we can introduce an SBI call to do that.

atishp04:
Hmm. I get your point about adding an additional SBI call.
My current implementation was clearing the pending bit in timer interrupt handler which probably might not be a very good idea in this case.

I am going to look for any possible alternate solution. Let me check how it is done in ARM.

aswaterman:
Cool, let us know how it goes...

aswaterman:
Sounds reasonable to me.  Thanks for the update.

atishp04:
Thank you :).
Closing the PR

PR #112, opened 2018-06-28
Libfdt support & local Timer node addition.

atishp04:
This PR adds support for libfdt and adds local timer node into the device tree. It is just a temporary patch only meant for linux capable boards. I hope future replacement of bbl will contain libfdt support inbuilt. Thus 1st patch will not be required afterwards.
That's why it is rebased on the commit id 5a0e3e55cab1d6f3462c77e852f16365c79f2c98 which was referred in freedom-u-sdk.

Please let me know if I should rebase these patches on top of some other commit.


atishp04:
FYI: I have reused code from glibc in util/bcopy.c & string.c. It was all under openBSD license. So I thought it would be okay to directly include. If not, please advise.

PR #114, opened 2018-07-26
Fix stubs-lp64.h/stubs-lp64d.h/stubs-ilp32.h/stubs-ilp32d.h not found

zongbox:
For now, we always compile files with -mabi option by using
soft-fp ABI. But there are some files use the header which in
toolchain, like unistd.h, stdint.h and so on. Finally, these
header files include other header files which existing depend
on ABI such as stubs-lp64.h or stubs-lp64d.h. So it causes the
header file not found when using the double-float toolchain with
soft-float ABI.

Fix up by compiling files with -march and -mabi options only if
there is specifying the --with-arch option at configure time.

If use the --with-arch option at configure time, that means your
toolchain has multi-lib support, so always building riscv-pk
by soft-float ABI is fine. Otherwise, we don't compile files with
specifying march and mabi options explicitly

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -54,14 +54,11 @@ endif
 # default is. The -with-abi is not necessary for this project. Unconditionally
 # compile it with a no-float ABI. i.e., ilp32 for 32-bit and lp64 for 64-bit.
 
-ifeq (@WITH_ARCH@,)
-  march := -march=$(shell @CC@ -v 2>&1 | \
-                    sed 's/ /\n/g' | grep with-arch | awk -F= '{print $$2}')
-else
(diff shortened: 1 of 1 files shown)


zongbox:
ping

zongbox:
@aswaterman  @palmer-dabbelt 
ping

damaxi:
I had the same problem before, it looks like that we expect march defined by default but even if I use one (not expert here) I had that problem with missing headers. Your patch helped me a lot with compiling pk.
I used Arch here is my output:
riscv64-linux-gnu-gcc: error: missing argument to '-march='

@palmer-dabbelt @aswaterman 
Please pull it

PR #117, opened 2018-08-06
Add some exception handling functions to s-mode to handle exception return from m-mode.

wxjstz:
Some exceptions will enter the M mode first, and will return to the S mode through the exceptions return after the M mode cannot be processed.

> Instruction access fault
> Load access fault
> Store/AMO access fault
> Illegal instruction
> misaligned load
> misaligned store
	
	
Some code is missing in the S-mode exception handling code (pk/handlers.c).
> Instruction access fault
> Load access fault
> Store/AMO access fault
> misaligned load

At least needs to be report error

### DIFF
--- a/pk/handlers.c
+++ b/pk/handlers.c
@@ -5,6 +5,24 @@
 #include "syscall.h"
 #include "mmap.h"
 
+static void handle_instruction_access_fault(trapframe_t *tf)
+{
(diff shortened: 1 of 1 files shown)


aswaterman:
Yes, this is much better than simply failing the assertion.

PR #122, opened 2018-08-24
Fix bug of mcall_set_timer

wxjstz:
mcall_set_timer is used to set the timed event. If only timecmp is set, there will be no timing effect.

### DIFF
--- a/machine/mtrap.c
+++ b/machine/mtrap.c
@@ -88,6 +88,7 @@ static uintptr_t mcall_shutdown()
 
 static uintptr_t mcall_set_timer(uint64_t when)
 {
+  *HLS()->time = 0;
   *HLS()->timecmp = when;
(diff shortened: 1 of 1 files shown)


aswaterman:
It was correct before.  The argument is not a delta; it is an absolute time.

wxjstz:
ok! Thank you! I know!!!

```c
static int next_event(unsigned long delta, struct clock_event_device *ce)
{
	/*
	 * time_init() allocates a timer for each CPU.  Since we're writing the
	 * timer comparison register here we can't allow the timers to cross
	 * harts.
	 */
	BUG_ON(ce != this_cpu_ptr(&riscv_clock_event));
	sbi_set_timer(get_cycles64() + delta);
	return 0;
}
```

PR #127, opened 2018-10-11
Add "--with-max-memory-bytes", to limit the used memory

palmer-dabbelt:
I have no idea if this works, but it does at least compile.

Signed-off-by: Palmer Dabbelt <palmer@dabbelt.com>


aswaterman (changes requested):
untested

PR #132, opened 2018-11-06
Don't require a RISC-V libc and crt when configuring

jrtc27:
We're always linking statically without a libc and crt files, but at configure time autoconf tries to link an executable, and will fail to do unless you have these available, so we should set LDFLAGS before performing any configure checks.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -114,7 +114,7 @@ COMPILE       := $(CC) -MMD -MP $(CFLAGS) \
 #  - LIBS    : Library flags (eg. -l)
 
 LD            := $(CC)
-LDFLAGS       := @LDFLAGS@ -nostartfiles -nostdlib -static $(LDFLAGS) $(march) $(mabi)
+LDFLAGS       := @LDFLAGS@ $(LDFLAGS) $(march) $(mabi)
 LIBS          := @LIBS@
 LINK          := $(LD) $(LDFLAGS)
 
--- a/configure
+++ b/configure
@@ -2043,6 +2043,7 @@ case $host_os in *\ *) host_os=`echo "$host_os" | sed 's/ /-/g'`;; esac
 # Checks for programs
 #-------------------------------------------------------------------------
 
+LDFLAGS="$LDFLAGS -nostartfiles -nostdlib -static"
 ac_ext=c
 ac_cpp='$CPP $CPPFLAGS'
 ac_compile='$CC -c $CFLAGS $CPPFLAGS conftest.$ac_ext >&5'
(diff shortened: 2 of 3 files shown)


davidlt:
This is causing configure errors on native build. Reverting this commit solves the problem.
```
[root@fedora-riscv build]# ../configure --enable-logo --with-payload=/boot/vmlinuz-5.0.0-0.rc2.git0.1.0.riscv64.fc30.riscv64
checking build system type... riscv64-unknown-linux-gnu
checking host system type... riscv64-unknown-linux-gnu
checking for gcc... gcc
checking whether the C compiler works... yes
checking for C compiler default output file name... a.out
checking for suffix of executables... 
checking whether we are cross compiling... configure: error: in `/root/rpmbuild/BUILD/riscv-pk/build':
configure: error: cannot run C compiled programs.
If you meant to cross compile, use `--host'.
See `config.log' for more details
```

```
[..]
configure:2391: checking whether the C compiler works
configure:2413: gcc    -nostartfiles -nostdlib -static conftest.c  >&5
/usr/bin/ld: warning: cannot find entry symbol _start; defaulting to 00000000000100d4
configure:2417: $? = 0
configure:2465: result: yes
configure:2468: checking for C compiler default output file name
configure:2470: result: a.out
configure:2476: checking for suffix of executables
configure:2483: gcc -o conftest    -nostartfiles -nostdlib -static conftest.c  >&5
/usr/bin/ld: warning: cannot find entry symbol _start; defaulting to 00000000000100d4
configure:2487: $? = 0
configure:2509: result: 
configure:2531: checking whether we are cross compiling
configure:2539: gcc -o conftest    -nostartfiles -nostdlib -static conftest.c  >&5
/usr/bin/ld: warning: cannot find entry symbol _start; defaulting to 00000000000100d4
/usr/bin/ld: /tmp/ccsr0hrL.o: in function `main':
conftest.c:(.text+0x14): undefined reference to `fopen'
/usr/bin/ld: conftest.c:(.text+0x24): undefined reference to `ferror'
/usr/bin/ld: conftest.c:(.text+0x34): undefined reference to `fclose'
collect2: error: ld returned 1 exit status
configure:2543: $? = 1
configure:2550: ./conftest
../configure: line 2552: ./conftest: No such file or directory
configure:2554: $? = 127
configure:2561: error: in `/root/rpmbuild/BUILD/riscv-pk/build':
configure:2563: error: cannot run C compiled programs.
If you meant to cross compile, use `--host'.
See `config.log' for more details
```

aswaterman:
Reverting.   @jrtc27 if you can provide a version of this patch that doesn't break the native build, we can merge it.  The temporary workaround is to use a toolchain that has libc/crt files, e.g., the newlib toolchain from riscv-gnu-toolchain.

PR #138, opened 2018-12-19
Use PMP to protect the loader from S-mode payloads

aswaterman:
The purpose of this isn't ironclad security; this loader is still written to assume the S-mode payload is trusted.  The intent is to catch OS bugs early: the OS should never be accessing the loader's memory.

### DIFF
--- a/bbl/bbl.c
+++ b/bbl/bbl.c
@@ -40,6 +40,52 @@ static void filter_dtb(uintptr_t source)
   filter_compat(dest, "riscv,debug-013");
 }
 
+static void protect_memory(void)
+{
+  // Check to see if up to four PMP registers are implemented.
+  // Ignore the illegal-instruction trap if PMPs aren't supported.
+  uintptr_t a0 = 0, a1 = 0, a2 = 0, a3 = 0, tmp, cfg;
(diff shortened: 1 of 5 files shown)


bbl/bbl.c
```diff
@@ -40,6 +40,52 @@ static void filter_dtb(uintptr_t source)
   filter_compat(dest, "riscv,debug-013");
 }
 
+static void protect_memory(void)
+{
+  // Check to see if up to four PMP registers are implemented.
+  // Ignore the illegal-instruction trap if PMPs aren't supported.
+  uintptr_t a0 = 0, a1 = 0, a2 = 0, a3 = 0, tmp, cfg;
+  asm volatile ("la %[tmp], 1f\n\t"
+                "csrrw %[tmp], mtvec, %[tmp]\n\t"
```
palmer-dabbelt: I think this is OK: it'll break if an m-mode interrupt comes in, but there shouldn't be any that can fire here because everything is delegated by hardware to S mode.

aswaterman: In general, bbl/pk assume that interrupts are disabled while in M-mode (which is the case).

PR #145, opened 2019-02-28
Stub out madvise syscall

luismarques:
This is needed for some programs linked with musl.


aswaterman:
IIRC it’s not safe to unconditionally return 0 for madvise, because contrary to its name, it provides more than advice. e.g. DONTNEED memory to be freed and zeroed immediately. Will musl barf on ENOSYS? Or maybe only return 0 for the ones that are simply advisory, and ENOSYS for everything else?

luismarques:
ENOSYS works for at least the issue that bothered me. I had a quick look at the musl codebase and I didn't find any usage for madvise that was obviously relevant and problematic for pk. Ready to pull.

aswaterman:
OK, let's do this for now, thanks!

Issue #105, opened 2018-05-15
undefined reference to `__stack_chk_guard'


ibrep:
Failed with Linking process.

[...]
riscv64-linux-gnu-gcc -Wl,--build-id=none -nostartfiles -nostdlib -static   -o pk pk.o -L.  -lpk  -lmachine  -lsoftfloat  -lutil -lgcc -T ../pk/pk.lds
pk.o: In function `rest_of_boot_loader':
pk.c:(.text+0x0): undefined reference to `__stack_chk_guard'
./libpk.a(file.o): In function `file_pwrite':
file.c:(.text+0x39c): undefined reference to `__stack_chk_guard'
./libpk.a(file.o): In function `.L0 ':
file.c:(.text+0x406): undefined reference to `__stack_chk_fail'
./libpk.a(elf.o): In function `load_elf':
elf.c:(.text+0x0): undefined reference to `__stack_chk_guard'
elf.c:(.text+0x1e8): undefined reference to `__stack_chk_guard'
elf.c:(.text+0x216): undefined reference to `__stack_chk_fail'
./libpk.a(console.o): In function `.L0 ':
console.c:(.text+0x4): undefined reference to `__stack_chk_guard'
console.c:(.text+0x50): undefined reference to `__stack_chk_fail'
./libpk.a(console.o): In function `.L7':
console.c:(.text+0x5c): undefined reference to `__stack_chk_guard'
./libpk.a(console.o): In function `.L0 ':
console.c:(.text+0x96): undefined reference to `__stack_chk_fail'
./libpk.a(console.o): In function `.L22':
console.c:(.text+0x150): undefined reference to `__stack_chk_guard'
./libpk.a(syscall.o): In function `sys_chdir':
syscall.c:(.text+0x750): undefined reference to `__stack_chk_guard'
./libpk.a(syscall.o): In function `sys_ftruncate':
syscall.c:(.text+0x7da): undefined reference to `__stack_chk_fail'
syscall.c:(.text+0x7e6): undefined reference to `__stack_chk_guard'
./libpk.a(syscall.o): In function `.L168':
syscall.c:(.text+0x898): undefined reference to `__stack_chk_fail'
./libmachine.a(mtrap.o): In function `.L47':
mtrap.c:(.text+0x16c): undefined reference to `__stack_chk_guard'
./libmachine.a(mtrap.o): In function `.L0 ':
mtrap.c:(.text+0x18e): undefined reference to `__vsnprintf_chk'
[...]

PR #150, opened 2019-03-23
Build fixes

lsgunth:
Two patches:

1) The first one fixes the build issues when using the compiler that's readily available in Ubuntu. This is the issue dismissed in #105. There's nothing wrong with this compiler except it enables stack protection and FORTIFY_SOURCE by default and we don't want these for riscv-pk. Adding flags to disable them allows us to use the compiler with no ill effects. The compiler readily available in Debian does not have this problem and is not affected by the flags.

2) Always create bbl.bin with objcopy which is typically used by real hardware (qemu likes the ELF bbl). This was done in the freedom-u-sdk, but using that SDK is a giant pain and building our kernels manually is much easier. So it's much prefered for this file to be created by riscv-pk as well.

Thanks,

Logan

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -102,7 +102,7 @@ VPATH := $(addprefix $(src_dir)/, $(sprojs_enabled))
 CC            := @CC@
 READELF       := @READELF@
 OBJCOPY       := @OBJCOPY@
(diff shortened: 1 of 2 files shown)


luismarques:
Isn't bbl deprecated in favor of opensbi?

lsgunth:
I don't know anything about that, but I don't think it's really relevant. It seems people are still using this project.

If bbl is deprecated, it's not well indicated anywhere.

PR #156, opened 2019-04-14
Update configure option for 32-bit

ak-fau:
configure script does not support --enable-32bit option since June 2018
(commit c3cf29a8f2f5f6a0b793bd1f24b083a759370a01). The right step
to enable 32-bit support in pk is to provide '--with-arch=rv32i' option to
configure.


aswaterman:
thanks!

PR #158, opened 2019-04-15
bbl/bbl.ac: use = instead of == for comparing equality

maxice8:
== is not defined in POSIX and will fail when configure
is called with shells that adhere strictly to POSIX like
dash

### DIFF
--- a/bbl/bbl.ac
+++ b/bbl/bbl.ac
@@ -1,7 +1,7 @@
 # See LICENSE for license details.
 
 AC_ARG_ENABLE([logo], AS_HELP_STRING([--enable-logo], [Enable boot logo]))
-AS_IF([test "x$enable_logo" == "xyes"], [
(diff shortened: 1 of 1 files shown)


maxice8:
These should be the last bits required for the configure script to be POSIX compliant

aswaterman:
Thanks for fixing our bashisms

maxice8:
On Mon, Apr 15, 2019 at 10:32:42AM -0700, Andrew Waterman wrote:
> Merged #158 into master.
> 
> -- 
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub:
> https://github.com/riscv/riscv-pk/pull/158#event-2277351294

I recommend the usage of shellcheck for scripts, it will report on any bashisms when dealing with files that end in .sh or have #!/bin/sh

PR #161, opened 2019-05-07
change MAX_ARGS from 64 to 1024

rink1969:
MAX_ARGS 64 too samll.
pk will report assert if program have long argument.

### DIFF
--- a/pk/pk.c
+++ b/pk/pk.c
@@ -29,7 +29,7 @@ static void handle_option(const char* s)
   }
 }
 
-#define MAX_ARGS 64
+#define MAX_ARGS 1024
 typedef union {
   uint64_t buf[MAX_ARGS];
   char* argv[MAX_ARGS];
@@ -143,7 +143,7 @@ static void run_loaded_program(size_t argc, char** argv, uintptr_t kstack_top)
 
 static void rest_of_boot_loader(uintptr_t kstack_top)
 {
-  arg_buf args;
+  static arg_buf args;
   size_t argc = parse_args(&args);
   if (!argc)
     panic("tell me what ELF to load!");


aswaterman:
8 KiB for this structure is kind of extreme. Do you really need 1024? Is 128 or 256 not sufficient?

rink1969:
Arguments paased as string, So 8KiB only means 1024 char.
I'm run program about Crypto. Length of Pubkey is 66 and length of signature is 140.

aswaterman:
I just tried running a program with a 66-char argument and a 140-char argument, and as I expected, it worked fine without increasing the buffer size.  Can you provide the test case?

rink1969:
My testcase
```
# spike pk multi-signature 96f4093cf179aaa369379402d74f70090fae11ec 02ebb3bcef8da55dc9884fae227ef9784625dc81eafd46c564d1045c7fa1cd9fa3 304502210090c8cc98ef64845836265db38c06a9a3b939c119e817542f9e8e0c9e2f5adf7c022029c9b0a22952a46da3b3fccea726e184625fa5228b98835adbd694b5bece9eaf 02a8a205c13f8da23a897b9cea857152efd898d8fa79bb5fc0219f2a89b132fd2a 30440220378b1b14a37ad8a4629fec7de213db83e149165a35ced69ab732fbc2dbfaf6fd02202e8e6b30b94fdbf9f067ea1f7f8f924e6eb82987b074f9a71c1a00e04ef3dc45 0241647ae455785b85cfb16d52d5619e7ee079b0b304a2c7fef1df0084fd4bcb3c 3045022100b9516ff7d839b7ca74c2662e1f0637e82705e29c575e194b2b8b758a3a91a03e022010438bb3ebb10ac55ff0e47f981b399b1e4c302a74855a579f4daa130818499a 
bbl loader
assertion failed @ 0x0000000080000518: r == 0
```
The program is verify multi-signature. So there are three groups pubkey and signature.
You can run this testcase without program.
Assert occurred before entry point of program.

Maybe 1024 is too big for current testcase.
But if I need verify more groups of pubkey and signature. The length of arguments are easy to bigger than 1024.
In fact, I have no idea about the exact value of MAX_ARGS.

Another point is pk will crash after increase MAX_ARGS.
Because the local var `arg_buf args` caused stack overflow.
So I add `static` to move it out of stack.

aswaterman:
I see.  MAX_ARGS is a bad name, what it really means is that there are MAX_ARGS * 8 bytes available for the total structure (512 B in this case).  Each argument character takes up 1 byte (including the null terminator), and each individual argument needs 8 bytes for a pointer.  So those args take up around 750 B.

I'll increase MAX_ARGS to 256, which allows up to 2 KiB of args--roughly 3x what this test case needs.  This is small enough to avoid the `static`, which is my main problem: adding more stuff to `.bss` increases loading time, which slows down RTL simulation a lot.

aswaterman:
I'll also improve the error message when the args get too big.

PR #162, opened 2019-06-05
Implement dup3 syscall

luismarques:
What it says on the tin.

### DIFF
--- a/pk/file.c
+++ b/pk/file.c
@@ -53,6 +53,20 @@ int file_dup(file_t* f)
   return -1;
 }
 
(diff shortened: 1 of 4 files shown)


pk/file.c
```diff
@@ -53,6 +53,16 @@ int file_dup(file_t* f)
   return -1;
 }
 
+int file_dup3(file_t* f, int newfd)
```
aswaterman: Need to check that newfd is >= 0 and < MAX_FDS. Looks good otherwise.

luismarques:
ping.

PR #163, opened 2019-06-11
Building BBL for the LiteX/Rocket SoC

gsomlo:
The LiteX SoC (https://github.com/enjoy-digital/litex) is now capable of hosting a 64-bit Rocket Chip as one of its CPU choices.

The enclosed patches allow BBL to be loaded by the LiteX/Rocket SoC:

1. LiteX uses a 64-bit Rocket variant built using `DefaultFPGAConfig` and `WithNSmallCores`, which doesn't support `U` mode and traps when `mcounteren` is accessed.
2. LiteX maps physical ram to `0x40000000`, so the hard-coded assumption of `0x80000000` needs to become an override-able default in pk/bbl.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -278,7 +278,7 @@ $$($(2)_install_prog_objs) : %.o : %.c
 	$(COMPILE) -c $$<
 
 $$($(2)_install_prog_exes) : % : %.o $$($(2)_prog_libnames)
-	$(LINK) -o $$@ $$< $$($(2)_prog_libarg) $(LIBS) -T $(src_dir)/$(2)/$(2).lds
+	$(LINK) -o $$@ $$< $$($(2)_prog_libarg) $(LIBS) -Wl,--defsym=MEM_START=@MEM_START@,-T,$(src_dir)/$(2)/$(2).lds
 
 $(2)_c_deps += $$($(2)_install_prog_deps)
 $(2)_junk += \
--- a/bbl/bbl.lds
+++ b/bbl/bbl.lds
@@ -12,7 +12,7 @@ SECTIONS
   /*--------------------------------------------------------------------*/
 
   /* Begining of code and text segment */
-  . = 0x80000000;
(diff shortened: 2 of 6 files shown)


aswaterman:
These changes look fine, but I'm curious what you are using BBL for if you don't have U-mode.

gsomlo:
On Tue, Jun 11, 2019 at 08:16:13AM -0700, Andrew Waterman wrote:
> These changes look fine, but I'm curious what you are using BBL for if you
> don't have U-mode.

I'm trying to use BBL to boot Linux on LiteX/Rocket. It may well end
up with me having to build a bigger version of 64-bit Rocket that
actually includes U-mode (BTW, would that have to be 'WithNBigCores()',
or can I get away with something less drastically larger?) to graft
onto the LiteX ecosystem.

But right now I'm just working on getting bits and pieces to compile,
until I hit the next hurdle (which is probably when I try to get BBL
to actually start Linux as a payload)... :)

aswaterman:
I think there are brand-new patches to support nommu linux, but for time being, you'll probably end up needing to use a core with virtual memory.  But there's nothing objectionable about the `mcounteren` patch, so we can merge it anyway.

gsomlo:
On Tue, Jun 11, 2019 at 08:37:13AM -0700, Andrew Waterman wrote:
> I think there are brand-new patches to support nommu linux,
> but for time being, you'll probably end up needing to use a
> core with virtual memory.

Thanks for merging. Yes, I positively want VM and U-mode; I don't
want (well, I do, but don't have room on my FPGA for) F and D, which
is why I used WithNSmallCores, without realizing that this will also
turn off VM and U-mode. I think Rocket needs an out-of-the-box
intermediate config where VM is enabled, but F and D are turned off.
I'll look into what that should look like config-wise, and where it
will fall in terms of size in between WithNSmallCores and WithNBigCores...

Thanks,
--G

PR #167, opened 2019-07-17
Fixes for booting with code above 2^32

kprovost:
I ran into two problems trying to boot on a system where memory is mapped above 0x100000000. This exposed bugs in the trap handler and illegal_inst_trap() when they compute the location of the handler function.

### DIFF
--- a/machine/emulation.c
+++ b/machine/emulation.c
@@ -145,6 +145,11 @@ void illegal_insn_trap(uintptr_t* regs, uintptr_t mcause, uintptr_t mepc)
   extern uint32_t illegal_insn_trap_table[];
   uint32_t* pf = (void*)illegal_insn_trap_table + (insn & 0x7c);
   emulation_func f = (emulation_func)(uintptr_t)*pf;
+#if __INTPTR_WIDTH__ == 64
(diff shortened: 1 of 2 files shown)


aswaterman:
Thanks for the bug fixes.  Unfortunately, these slow things down when the trap table is in the right place (and the performance of this code does matter to some extent..).  Let me see if I can find a way to preserve both your bug fix and the performance.

kprovost:
Thanks for the quick response. Let me know if there's anything I can do to help. I can reproduce the problem easily in qemu.

aswaterman:
Thanks.  I'll get you to test a patch in a sec.  I'm just going to convert these tables to use pointer-sized entries.  The extra data memory footprint shouldn't affect performance in practice, since there's usually huge temporal locality in accessing these tables.

aswaterman:
@kprovost Can you give this a shot? https://github.com/riscv/riscv-pk/pull/168

kprovost:
With #168 I can also boot my system, so yes, that's good too.

aswaterman:
Cool.  I'm going to merge that one.  Thanks for raising this issue.

PR #168, opened 2019-07-17
Fix bbl/pk living above 4 GiB

aswaterman:
This is an alternative to #167


yzt000000:
the current version is support boot from above 4GB memory space? why I always goto poweroff when boot from 0x100000000. 
did I missing some important item

PR #171, opened 2019-08-12
allow additional ns16550a config via device tree

ipatix:
This commit makes bbl read some additional fields from
the device tree if it detects an ns16550a:

- reg-shift
- reg-offset
- clock-frequency

For explanation of these check out the Linux Kernel doc:
https://www.kernel.org/doc/Documentation/devicetree/bindings/serial/8250.txt

In particular this allows the Xilinx AXI UART 16550 to act
as serial console with bbl and the Linux early boot console.

This also fixes a bug in which bbl will ignore any other than the first
"compatible" string when iterating over the nodes.
Previously this line would not have worked:

compatible = "xlnx,xps-uart16550-2.00.a", "ns16550a";

Before bbl would have just checked the first field instead of checking
all strings in the list.

### DIFF
--- a/machine/fdt.c
+++ b/machine/fdt.c
@@ -126,6 +126,11 @@ const uint32_t *fdt_get_size(const struct fdt_scan_node *node, const uint32_t *v
   return value;
 }
 
+uint32_t fdt_get_value(const struct fdt_scan_prop *prop, uint32_t index)
+{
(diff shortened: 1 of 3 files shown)


aswaterman:
Looks alright to me, but the maintainer of this code (@palmer-dabbelt) should weigh in.

PR #173, opened 2019-08-21
Try to work with linux kernel specified compat strings

tmagik:
Work with the new linux 'compatible' strings for uart


jrtc27:
It's currently `compatible = "sifive,fu540-c000-uart", "sifive,uart0";`, so after #202 it should now pick up the `sifive,uart0` in there again. I don't see a reason to match on the more-specific one given the driver doesn't currently do anything that doesn't apply to the generic SiFive UART; the whole point of having `compatible` be a list is so that generic drivers only need to match the generic thing, with the more-specific entry matched only when necessary.

Issue #160, opened 2019-04-22
DTS mmu-type mismatch with spike when using RV32


LwpDaRren:
when using RV32 instruction mode

riscv-pk/machine/fdt.c:649
`if (!strcmp(filter->mmu_type, "riscv,sv39")) return false;`

mismatch with the DTB generated by Spike
riscv-isa-sim/riscv/dts.cc:35
`mmu-type = \"riscv," << (procs[i]->get_max_xlen() <= 32 ? "sv32" : "sv48") << "\";\n"`

which will cause infinite-loop dead and print 

> hart_filter_mask saw unknown hart type: status="riscv,sv32", mmu_type="riscv,sv39"

modify "sv39" with "sv32" everything seems Okay
`if (!strcmp(filter->mmu_type, "riscv,sv32")) return false;`

PR #177, opened 2019-10-02
fdt: allow mmu type "riscv,32" on rv32 systems

gktrk:
SV32 is presented in RISC-V Privileged Architecture Manual (version
20190608-Priv-MSU-Ratified) Section 4.3 for RV32 systems. However, BBL
responds to sv32 with:

  hart_filter_mask saw unknown hart type: status="okay", mmu_type="riscv,sv32"

and hangs.

This patch is adopted from the original 'riscv-pk.diff' patch written
by Fabrice Bellard, distributed as part of the following tarball:
https://bellard.org/tinyemu/diskimage-linux-riscv-2018-09-23.tar.gz

Closes: https://github.com/riscv/riscv-pk/issues/160

### DIFF
--- a/machine/fdt.c
+++ b/machine/fdt.c
@@ -651,8 +651,12 @@ static bool hart_filter_mask(const struct hart_filter *filter)
 {
   if (filter->mmu_type == NULL) return true;
   if (strcmp(filter->status, "okay")) return true;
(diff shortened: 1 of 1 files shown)


gktrk:
Ping.

PR #182, opened 2019-11-06
Support a subset of 16750 functionality, and improve baud rate selection

jrrk2:
A non-default baud rate cannot be used. This patch fixes the problem in a way compatible to Linux. Also support a subset of 16750 functionality in the same file.


aswaterman:
LGTM

PR #185, opened 2020-01-31
Init mstatus for vector extension

kito-cheng:
spike will check mstatus.vs at https://github.com/riscv/riscv-isa-sim/commit/4051af5ce64a78c2d032ba7b0452b8bb929ac281 , so we must init mstatus.vs if vector extension enabled.


aswaterman:
I thought I already did this in the most recent two commits on master.

kito-cheng:
Oh seems like I forgot rebase to master and resolve the problem again.

aswaterman:
Cool.

PR #186, opened 2020-02-02
Support --without-payload for OpenSBI fw_jump-style booting

jrtc27:
We expect the firmware to load the external payload at the second
megapage, and that there is space to put the filtered FDT at 0x2200000
past the start of memory. With a default MEM_START of 0x80000000, this
matches the standard OpenSBI values for FW_JUMP_ADDR and
FW_JUMP_FDT_ADDR of 0x80400000/0x80200000 (RV32/RV64) and 0x82200000
respectively, so payloads linked for one should work with the other.


jrtc27:
Ping?

aswaterman:
Lgtm.

PR #188, opened 2020-02-02
Support manually zeroing out BSS when booting

jrtc27:
Some ELF loaders, in particular gdb's load command for dynamically
loading files into memory, which is often used to load binaries onto
FPGAs over JTAG, do not zero out BSS, leaving the memory in whatever
state it was previously in. Thus, introduce a new --enable-zero-bss
configure flag, which will include code to zero out BSS when booting.


aswaterman:
Lgtm.

PR #192, opened 2020-03-16
Fix CPI calcualtion for -s option

davidmetz:
This attempt at rounding leads to wrong results, for example:
314689951 cycles
314690101 instructions
0.90 CPI
With my change results in:
314689951 cycles
314690101 instructions
0.99 CPI
I think this was supposed to be part of rounding behaviour but it doesn't work if only the final digit does it and there is no carry...
Instead I changed it to truncate after the second digit


aswaterman:
Would be better to round, but I agree truncation is better than producing a completely incorrect result :)

PR #202, opened 2020-06-07
Consistently use fdt_string_list for FDT compatible property

jrtc27:
QEMU's finisher is "sifive,test1\0sifive,test0\0syscon" so we fail to
detect it currently. Instead, search the entire list, and for
completeness do the same with the HTIF and SiFive UART drivers.


aswaterman:
LGTM, thanks.

PR #205, opened 2020-07-02
Don't perform 64-bit accesses to the PLIC

arichardson:
Recent QEMU will fault for 8-byte accesses. Use a uint32_t instead of
uintptr_t to avoid those problems.


jrtc27:
Ping. This is critical for BBL to work on the latest QEMU 5.1.0-rcX releases, and will be needed for the upcoming 5.1.0 proper release.

arunthomas:
+1, we've also run into this issue on QEMU 5.1-rcX. I can confirm the PR resolves the issue.

PR #211, opened 2020-08-08
Make htif_poweroff thread-safe

zhemao:
This fixes an issue we were seeing in which a dual-core BOOM simulation wasn't shutting down properly. I believe the root cause was that the core calling htif_poweroff was getting into a data race with the other core calling htif_console_getchar. Having htif_poweroff take the htif lock and use __set_tohost to write the tohost register seems to solve the issue.


aswaterman:
Makes perfect sense.  Thanks, @zhemao and @timsnyder-siv

PR #216, opened 2020-10-10
Fix handling of compressed breakpoints

anthony-coulter:


aswaterman:
thanks!

aswaterman:
I think the better fix is to just kill the program, rather than skipping over the breakpoint.  Then we don't need the epc increment.

PR #217, opened 2020-10-28
Disable device tree filter when load a dts from file

Phantom1003:


aswaterman:
Does this change the default behavior?  I think adding the option is fine, but we should keep the default behavior consistent.

Phantom1003:
This filter is used to keep compatibility four years ago, recently I try to use bbl as sbi to boot linux, and I found there is a panic during booting, it says irq 0xffffffff is too large for interrupt controller. I find these old filter will remove clint, and change SEIP to -1 etc. These operation is not neccessary, and when we specify a dts, we hope bbl directly help us pass it to kernel, instead of changing something ...

aswaterman:
Got it.  I think this is OK.

aswaterman:
This breaks the build when `--with-dts` is not specified, so I am reverting it.  Please resubmit the PR in a way that works both with and without the argument.

Phantom1003:
I'm very sorry for the trouble I've caused by my oversight.This is the first time I've submitted code to the community, I simply tested it in my own environment, I used macros to maintain compatibility, but forgot the dependencies in make, sorry about that again, I'll make sure not to make this mistake again!
New request is at #219

aswaterman:
No worries, thanks for the quick fix. I’ll review the new PR shortly.

PR #219, opened 2020-10-31
Disable device tree filter when load a dts from file

Phantom1003:
In past versions, bbl would filter the device tree to maintain compatibility. Now with the development of software systems, most of the device tree properties are supported by linux and we don't need this filtering anymore. bbl is just a simple bootloader, although in practice we should use opensbi, but in order to enable bbl to boot linux on more complex devices, I have added the --with-dts option and when this option is enabled, we can specify a device tree file to describe the onboard resources, and properties will not be filtered out by default.


aswaterman:
Thanks!

PR #222, opened 2020-11-17
Remove use of copy_stat() from sys_fstatat

### LINKED ISSUE #221: Double conversion of struct stat
When responding to the syscall `SYS_stat` (or related variants), pk calls the function `copy_stat()` to convert the `struct frontend_stat` provided by spike into a regular `struct stat`.  However, after returning from the syscall, newlib calls the function `_conv_stat()`, which _also_ converts to a regular `struct stat`, this time from a `struct kernel_stat` (which looks identical to the `struct frontend_stat`).  So the stat struct gets converted twice, with a garbage result.

I don't know whether it's the pk or newlib which should stop converting, I guess pk should use the same output format as the linux kernel does?  (Edit:  The riscv native linux kernel, I mean.)

zeldin:
Fixes #221 

I have tested this with both a newlib and a glibc payload program.


aswaterman:
Thanks for the detective work on this one.

PR #224, opened 2020-11-22
Fix emulation of misaligned access on big endian target

zeldin:


aswaterman:
Thanks.  I'm not surprised this code was full of endianness assumptions.

PR #228, opened 2020-11-29
Remove use of copy_stat() from sys_fstat and sys_lstat

zeldin:
I found some more places where the stat struct was mangled...  :smile_cat:

### DIFF
--- a/pk/syscall.c
+++ b/pk/syscall.c
@@ -132,7 +132,9 @@ int sys_fstat(int fd, void* st)
 
   if (f)
   {
-    r = file_stat(f, st);
+    struct frontend_stat buf;
+    r = frontend_syscall(SYS_fstat, f->kfd, va2pa(&buf), 0, 0, 0, 0, 0);
+    memcpy(st, &buf, sizeof(buf));
     file_decref(f);
   }
 
@@ -215,7 +217,7 @@ long sys_lstat(const char* name, void* st)
   struct frontend_stat buf;
   size_t name_size = strlen(name)+1;
   long ret = frontend_syscall(SYS_lstat, va2pa(name), name_size, va2pa(&buf), 0, 0, 0, 0);
-  copy_stat(st, &buf);
+  memcpy(st, &buf, sizeof(buf));
   return ret;
 }
 


aswaterman:
@zeldin How is this working on RV32, btw?

zeldin:
@aswaterman It looks like at least glibc wants to use a `stat64` syscall in 32-bit mode.  When testing though I find that 32-bit pk can't run even a simple hello world program, so I need to investigate that first.  I also found that the `riscv32-unknown-elf-gcc` built by `riscv-gnu-toolchain` actually generated 64 bit code.  Curiouser and curiouser...

aswaterman:
Thanks for volunteering to investigate further. Probably best to start with glibc.

zeldin:
So the `riscv-gnu-toolchain` issue turned out to be PEBKAC.  I incorrectly assumed that `make XLEN=32` would give me a 32-bit toolchain, but actually `WITH_ARCH` and `WITH_ABI` must be overridden as well.  :smile_cat:

zeldin:
And the problem with running applications under the 32-bit `pk` seems to be related to my use of the `-s` flag; without it the applications run.

So the result of the stat test is:
* newlib: Both stat and fstat work correctly
* glibc: The program is terminated with the message
```
FATAL: kernel too old
bad syscall #178
```
before even getting into `main`.  Syscall 178 seems to be `gettid`...

zeldin:
Note that I have a newer version of glibc for 32-bit than I have for 64-bit (the older glibc does not support RV32).

aswaterman:
Thanks. We can figure out the glibc matter once someone finds time to update pk to support enough new syscalls and report a newer kernel version. If that someone is you, great :P but it might or might not end up being a tar pit, so I’d understand if you don’t want to.

zeldin:
I think that the use of syscall 178 is actually part of the implementation of `abort()` (the syscall it tries next after stubbing out `gettid` is `tgkill`), so that might be a red herring.  But yeah, kernel version reported probably needs a bump at least.

aswaterman:
Ah yeah, that makes sense.

zeldin:
Bumping reported kernel version to 5.4.0 (see [here](https://github.com/bminor/glibc/blob/master/sysdeps/unix/sysv/linux/riscv/configure.ac#L16-L18)) allows 32-bit glibc programs to run, but the `stat` and `fstat` glibc functions need the `statx` syscall...

PR #230, opened 2020-12-15
Add support for the UART interface on the LiteX SoC

gsomlo:
[LiteX](https://github.com/enjoy-digital/litex) is a FOSS SoC written in Python/Migen, which supports a wide range of CPUs and peripherals, and several of its configurations are capable of running Linux.

I've managed to successfully boot *almost* unmodified upstream Linux (with the only exception being LiteX "gateware" specific device drivers) on a LiteX SoC configured with [RocketChip](https://github.com/chipsalliance/rocket-chip), with the help of BBL (*for the full details see [here](http://www.contrib.andrew.cmu.edu/~somlo/BTCP/#sec_3)).

The attached PR adds support for LiteX's UART interface to BBL. Thanks for considering!

### DIFF
--- a/machine/machine.mk.in
+++ b/machine/machine.mk.in
@@ -15,6 +15,7 @@ machine_hdrs = \
   mtrap.h \
   uart.h \
   uart16550.h \
+  uart_litex.h \
   finisher.h \
   unprivileged_memory.h \
   vm.h \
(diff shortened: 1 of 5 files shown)


aswaterman:
lgtm

PR #234, opened 2021-03-25
Add `statx` syscall

huaixv:
The `statx` syscall is required to run some 32bit programs on Spike and `riscv-pk`, so i tried to add it to the proxy kernel. :)

Please let me know if there exist some problems with the code.

### DIFF
--- a/pk/frontend.h
+++ b/pk/frontend.h
@@ -32,4 +32,6 @@ struct frontend_stat {
   uint32_t __unused5;
 };
 
+#define FRONTEND_STATX_SIZE 256
+
 #endif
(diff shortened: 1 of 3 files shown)


pk/frontend.h
```diff
@@ -32,4 +32,36 @@ struct frontend_stat {
...
+    struct frontend_statx_timestamp mtime;
+    uint32_t rdev_major;
+    uint32_t rdev_minor;
+    uint32_t dev_major;
+    uint32_t dev_minor;
+    uint64_t mnt_id;
+    uint64_t __spare2;
+    uint64_t __spare3[12];
+};
+
```
aswaterman, line 66: I think you can delete all of this and just have a macro `FRONTEND_STATX_SIZE`.  The contents of the structure don't really matter; pk only needs to know how big it is.

huaixv: Fixed. :)

PR #235, opened 2021-03-25
LiteX UART: fix compatible property name

gsomlo:
The upstream LiteX project defaults to "litex,liteuart" as the value
for the "compatible" property of the UART DT node.

Signed-off-by: Gabriel Somlo <gsomlo@gmail.com>

### DIFF
--- a/machine/uart_litex.c
+++ b/machine/uart_litex.c
@@ -45,7 +45,8 @@ static void uart_litex_prop(const struct fdt_scan_prop *prop, void *extra)
 {
     struct uart_litex_scan *scan = (struct uart_litex_scan *)extra;
     if (!strcmp(prop->name, "compatible") &&
-        !strcmp((const char *)prop->value, "litex,uart0")) {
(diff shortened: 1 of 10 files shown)


aswaterman:
But I'm assuming there must be some implementations that still report `litex,uart0`, right?  Assuming the same driver works for both, should we not accept both strings?

gsomlo:
On Thu, Mar 25, 2021 at 03:34:48PM -0700, Andrew Waterman wrote:
> But I'm assuming there must be some implementations that still report
> litex,uart0, right? Assuming the same driver works for both, should we not
> accept both strings?

I'm not aware of any pre-existing use cases where the .dts file wasn't
embedded into BBL at build time (i.e., via `--with-dts=foo.dts`), but
we can definitely exercise an abundance of caution and accept both.
I'll send an update shortly.

gsomlo:
Replaced by PR #237 

Not sure where all the merge conflicts came from, so I re-applied my commit on top of the new upstream master.

Issue #239, opened 2021-04-28
Calls to __clear_cache() require pointer arguments


cmuellner:
When compiling with a recent version of GCC, we run into the following error:

```
      ../pk/pk.c: In function 'run_loaded_program.constprop':
      ../pk/pk.c:177:3: error: both arguments to '__builtin___clear_cache'
      must be pointers
        177 |   __clear_cache(0, 0);
            |   ^~~~~~~~~~~~~~~~~~~
```

The error comes from the fact, that we pass numeric values as arguments to the invocation of [__clear_cache](https://github.com/riscv/riscv-pk/blob/master/pk/pk.c#L177).

Using `(void*)0` or `NULL` instead of `0` does not work either.

My understanding of the code is, that we know that we can only clear the complete instruction cache and therefore we don't care about the arguments. Is this assumption correct?

This error has been introduced by the GCC commit [c05ece9 in Dec 2020](https://gcc.gnu.org/git/?p=gcc.git;a=commit;h=c05ece92c6153289fd6055e31e791e59b8ac412).
I have created a GCC bug report ([PR100316](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=100316)) as well.

cmuellner:
Ping for @aswaterman as he is the author of the affected piece of code.

aswaterman:
I think we can replace this one with an asm volatile fence.i.  If you make such a PR, I’ll merge it.

cmuellner:
The required code was already there. Just needed to be called.
Thanks!

aswaterman:
Resolved by #240.

PR #240, opened 2021-04-28
pk: Fix __clear_cache() compilation issue with recent compilers

cmuellner:
Using recent compilers we get the following error message:

  ../pk/pk.c: In function 'run_loaded_program.constprop':
  ../pk/pk.c:177:3: error: both arguments to '__builtin___clear_cache'
  must be pointers
    177 |   __clear_cache(0, 0);
        |   ^~~~~~~~~~~~~~~~~~~

Let's use the existing function __riscv_flush_icache(),
give it a header with a prototype and use it to
emits the FENCE.I instruction directly.

See #239

Suggested-by: Andrew Waterman <andrew@sifive.com>
Signed-off-by: Christoph Muellner <cmuellner@linux.com>

### DIFF
--- a/machine/flush_icache.h
+++ b/machine/flush_icache.h
@@ -1,0 +1,8 @@
+// See LICENSE for license details.
+
+#ifndef _RISCV_FLUSH_ICACHE_H
+#define _RISCV_FLUSH_ICACHE_H
+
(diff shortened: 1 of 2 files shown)


aswaterman:
Forgot about that function. LGTM!

PR #244, opened 2021-05-06
machine: manually perform assembler relaxation

compnerd:
This is an equivalent rewrite of the existing code.  When building with
gas, the `bltu` would implicitly get relaxed to the `bgeu` + `j`.  This
relaxation is required as the `init_other_hart` is not guaranteed to be
addressable in 12-bits.  When building with the LLVM IAS instead of gas
we fail to link as the branch is not relaxed.  This change enables LLVM
to build and link this code with the LLVM IAS and lld.

### DIFF
--- a/machine/mentry.S
+++ b/machine/mentry.S
@@ -312,7 +312,9 @@ do_reset:
   # make sure our hart id is within a valid range
   fence
   li a2, MAX_HARTS
-  bltu a3, a2, init_other_hart
+  bgeu a3, a2, .Lcont
+  j init_other_hart
+.Lcont:
 #endif
   wfi
   j .LmultiHart


aswaterman:
I don't have a problem with this patch, but I also don't think this is the end of the story.  LLVM's assembler seems deficient in this respect.  It won't be able to assemble code emitted by GCC unless it fixes this limitation.

compnerd:
I'm afraid that you are correct about that.  The LLVM assembler does need to be enhanced to support this relocation.  In general, use of LLVM currently requires that you build your code with `-mno-relax`, which has no impact on assembly though.

jrtc27:
I'd argue GNU as is wrong here. If I write `bltu` I should get `bltu` not `bgeu`+`j`. RISC-V GNU as is the outlier here, on other architectures this kind of code is not accepted.

PR #245, opened 2021-05-07
machine: fix a case of undefined behaviour with SP handling

compnerd:
The use of `asm` for register aliasing is supported in two different
contexts:
- local variables (including GNU expression statements) where it may
  only be used for specifying registers for input and output operands to
  extended `asm` syntax.

  c.f. https://gcc.gnu.org/onlinedocs/gcc/Local-Register-Variables.html#Local-Register-Variables

- global variables where it may be used to observe the contents of a
  register.

  c.f. https://gcc.gnu.org/onlinedocs/gcc/Global-Register-Variables.html#Global-Register-Variables

The two options here is to either to hoist the variable out into a
global variable, but then it should not be in a header due to fears of
ODR in case the optimizer does not inline it away, and thus becomes a
bit more tricky.  The alternative that this change actually adopts is to
explicitly use a move to copy the value out via the GNU extended
assembly syntax.

With this change, it is now possible to build the Proxy Kernel
completely with clang/LLVM and link with LLD.  The generated kernel also
runs under SPIKE and behaves as expected in a simple smoke test (without
any executable prints the expected message, and runs a trivial RVV
example).

### DIFF
--- a/machine/mtrap.h
+++ b/machine/mtrap.h
@@ -48,9 +48,14 @@ typedef struct {
   volatile uint32_t* plic_s_ie;
 } hls_t;
 
-#define MACHINE_STACK_TOP() ({ \
-  register uintptr_t sp asm ("sp"); \
-  (void*)((sp + RISCV_PGSIZE) & -RISCV_PGSIZE); })
+#define STACK_POINTER() ({            \
+  uintptr_t __sp;                     \
+  __asm__("mv %0, sp" : "=r"(__sp));  \
+  __sp;                               \
+})
+
+#define MACHINE_STACK_TOP()           \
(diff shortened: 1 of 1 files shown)


aswaterman:
Cool.

jrtc27:
This should have used `__builtin_frame_address(0)` as we did downstream and failed to upstream (https://github.com/CTSRD-CHERI/riscv-pk/commit/595d003bb1ea4134caefb930e48a6e5135477489). It avoids the use of opaque inline assembly and is more portable (since for CHERI the stack pointer is a capability and thus in `csp` not `sp`).

aswaterman:
Ah, nice.

PR #246, opened 2021-05-07
machine: correct some additional cases of UB

compnerd:
Use of asm aliased register variables in local scope can only be used
for extended assembly parameters.  This changes the few instances of
this in the floating point emulation to use the GNU extended assembly
syntax to access the `tp` register.  This ensures that we do not rely on
undefined behaviour.  This was uncovered when building the Proxy kernel
with clang and LLVM.

### DIFF
--- a/machine/fp_emulation.h
+++ b/machine/fp_emulation.h
@@ -47,14 +47,22 @@
   else tp = GET_RM(insn); \
   asm volatile ("":"+r"(tp)); })
 # define softfloat_raiseFlags(which) set_csr(fflags, which)
-# define softfloat_roundingMode ({ register int tp asm("tp"); tp; })
+# define softfloat_roundingMode ({    \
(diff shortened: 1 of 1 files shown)


jrtc27:
The less-invasive fix that will be slightly more performant (allows eliding the `mv`) is to just declare tp at global scope; this is what we did downstream to build with Clang and apparently never upstreamed.

jrtc27:
(https://github.com/CTSRD-CHERI/riscv-pk/commit/0d1fdc2e24b7b6247a55d24c13ae85dca7f45695, though that should probably also be `static`)

aswaterman:
Good point, I cherry-picked (CHERI-picked?) this alternative fix.

jrtc27:
> Good point, I cherry-picked (CHERI-picked?) this alternative fix.

:) as you can imagine we have fun with such puns...

PR #247, opened 2021-06-12
Set desired endianness at boot time

zeldin:


aswaterman:
This is a regression. Most RV32 cores don’t have the mstatush register, and so this will break them.

zeldin:
Hm, so how do you find out whether or not the register exists?  Is the only way trying to access it and handle the trap?

zeldin:
Is it ok to assume mstatush exists if pk is compiled as big endian?  I'm guessing all those cores which lack mstatush are little-endian only...
It's really only in that case where mstatush strictly needs to be written; if the core has configurable endianness it's supposed to go LE on reset, and if it's big endian only then writing mstatush will not do anything even if it exists...
Or maybe the problem then would be SBE (if it exists), since only MBE is required to be cleared on reset IIRC.

aswaterman:
Yes, @zeldin, I agree that’s a safe assumption, since the feature didn’t exist until the mstatush CSR also existed. Agreed @jhauser-us?

zeldin:
I updated the code to only write `mstatush` when `__BYTE_ORDER__ == __ORDER_BIG_ENDIAN__`.  I also removed the clear of `MBE` on RV64 little endian, for consistency.

aswaterman:
This version looks good to me; thanks, @zeldin.

machine/minit.c
```diff
@@ -24,6 +24,19 @@ void* kernel_end;
 static void mstatus_init()
 {
   uintptr_t mstatus = 0;
+#if __riscv_xlen == 32
+  uint32_t mstatush = 0;
+#endif
```
jrtc27, line 29: This presumably gives unused variable warnings for a little-endian RV32 build

jrtc27: And if asserts are disabled? Why not just ifdef it out here for little-endian?

zeldin: It just seems not so clean to spread out the assumption that the RV32 register MSTATUSH can only be safely written on big endian systems (which is kind of a workaround in the first place) to multiple locations.  But that's just my personal view.

PR #263, opened 2022-02-06
Untrack auto-generated files

ZenithalHourlyRate:
./configure and other config files
are machine dependent hence should
not be tracked by git.

As for aclocal.m4, it was not generated
by aclocal, instead it was copied from
mcppbs repo thus should be put to m4/

Unfortunately, we do not have Makefile.am
thus we can not untrack Makefile.in now

Reference: https://github.com/github/gitignore/blob/main/Autotools.gitignore


aswaterman:
Our convention in these repos is to commit these files to unburden users from needing the right versions of all of these extra tools.  So, I can't accept this PR.

aswaterman (changes requested):
Our convention in these repos is to commit these files to unburden users from needing the right versions of all of these extra tools. So, I can't accept this PR.

sequencer:
all distros supports the autotools by default, maybe this only needs some modification to the build script? Another possibility is porting repositories to modern build systems(like CMake, Meson).

aswaterman:
This isn't a problem that needs fixing.

sequencer:
I see...

aswaterman:
Yes, we are aware of the downsides of this approach. Please bear in mind that we are balancing multiple competing concerns here, and that we aren’t alone in making this decision. Consider another much more prominent pair of projects that does the same thing: https://github.com/bminor/binutils-gdb/blob/master/configure https://github.com/gcc-mirror/gcc/blob/master/configure

PR #264, opened 2022-02-06
No explicit -lgcc

ZenithalHourlyRate:
For Clang/LLVM user with NO cross gcc installed,
there is no libgcc available; At this time user
can manually export LIBS=-lclang-rt.

For cross gcc user, -lgcc is usually added by the
compiler thus there is no need to add this flag
in config file.
Only when -nostdlib/-nodefaultlibs is specified
should we add -lgcc, which is not the case here.

This compiles on a machine with cross gcc installed and another machine with only clang. However, only rv64 is tested.

Note: ./configure not re-generated w.r.t #263

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -96,7 +96,7 @@ LDFLAGS="$LDFLAGS -Wl,--build-id=none"
 
 AC_SUBST(CFLAGS)
 AC_SUBST(LDFLAGS)
-AC_SUBST([LIBS], ["-lgcc"])
+AC_SUBST(LIBS)
 AC_SUBST(WITH_ARCH)
 AC_SUBST(host_alias)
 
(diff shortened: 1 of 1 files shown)


aswaterman (changes requested):
This will break the build for target ISAs that require `ligbcc` support for e.g. int64 arithmetic.

A compromise might be to search for `-lgcc` and link it if it exists.

aswaterman:
Thanks!

aswaterman:
Incidentally, none of my `libgcc`s have `__ashlsi3`--which makes sense, since all RISC-V base ISAs have an instruction for left-shifting by 32 bits.  So this doesn't work for me.

But all of my `libgcc`s have `__divdc3` (division of `double _Complex`), and ISA support for that operation is not forthcoming, so I'll use that.

aswaterman:
Nevermind, that doesn't work for unrelated reasons.  But I think searching for both `__ashldi3` and `__ashlti3` does work (the former suffices for RV32 and the latter for RV64).

aswaterman:
Manually merged via c7e75bf944957e58f16531eb6b8d118a92069346

PR #265, opened 2022-02-06
Eliminate Wunknown-pragmas for Clang/LLVM

ZenithalHourlyRate:
With Clang/LLVM,
```
/home/zenithal/T/playground/dependencies/riscv-pk/util/string.c:9:13: error: unknown pragma ignored [-Werror,-Wunknown-pragmas]
#pragma GCC optimize ("no-tree-loop-distribute-patterns")
            ^
1 error generated.
```

### DIFF
--- a/util/string.c
+++ b/util/string.c
@@ -4,7 +4,7 @@
 #include <stdint.h>
 #include <ctype.h>
 
-#ifdef __GNUC__
+#if defined(__GNUC__) && !defined(__llvm__)
 // Don't let GCC pattern-match these functions' bodies into self-calls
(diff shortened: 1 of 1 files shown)


aswaterman:
Thanks!

util/string.c
```diff
@@ -4,7 +4,7 @@
 #include <stdint.h>
 #include <ctype.h>
 
-#ifdef __GNUC__
+#if defined(__GNUC__) && !defined(__llvm__)
```
jrtc27, line 7: Why `__llvm__` rather than `__clang__`? The latter is the normal thing to use...

PR #266, opened 2022-02-06
Eliminate Wuninitialized for Clang/LLVM

ZenithalHourlyRate:
```
/home/zenithal/T/playground/dependencies/riscv-pk/pk/console.c:68:43: error: variable 'ra' is uninitialized when used here [-Werror,-Wuninitialized]
  do_panic("assertion failed @ %p: %s\n", ra, s);
                                          ^~
/home/zenithal/T/playground/dependencies/riscv-pk/pk/console.c:67:24: note: initialize the variable 'ra' to silence this warning
  register uintptr_t ra asm ("ra");
                       ^
                        = 0
1 error generated.
```

Though this seems to be the fault of Clang.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -91,7 +91,7 @@ AS_IF([test "x$enable_print_device_tree" = "xyes"], [
   AC_DEFINE([PK_PRINT_DEVICE_TREE],,[Define if the DTS is to be displayed])
 ])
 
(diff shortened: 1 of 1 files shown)


aswaterman:
Workaround seems fine.

MaskRay:
Note this paragraph on https://gcc.gnu.org/onlinedocs/gcc/Local-Register-Variables.html#Local-Register-Variables

> Defining a register variable does not reserve the register.

`register uintptr_t ra asm ("ra");` not used as an input into an asm statement does not guarantee the value.
Consider `__builtin_return_address()`.

PR #280, opened 2022-06-27
add support for rve

liweiwei90:
This PR adds support for rve, if rve is enabled: 
* disable the handling of upper registers
* change the syscall number to register t0

### DIFF
--- a/dummy_payload/dummy_entry.S
+++ b/dummy_payload/dummy_entry.S
@@ -9,13 +9,21 @@ _start:
 1:
   lbu a0, (s0)
   beqz a0, 1f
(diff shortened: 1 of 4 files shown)


aswaterman:
Is there an RVE ABI that specifies the syscall number comes in `t0`?  It is a reasonable choice, but I want to make sure it is written down somewhere before adding software support for it.

aswaterman:
OK, I propose we leave this PR open until there is a written RV32E syscall ABI.

liweiwei90:
> OK, I propose we leave this PR open until there is a written RV32E syscall ABI.

OK.

PR #288, opened 2023-01-05
Regenerated the configure file with autoconf 2.71

langly:
Regenerated the configure file using autoconf 2.71 to avoid the Syntax error problem when running on OSX/AArch64.


aswaterman:
Hi @langly!!

PR #300, opened 2023-07-07
Add ENOSYS stub for syscall 278, getrandom

egouriou-rivos:
Binaries built with a recent LLVM and current riscv-gnu-toolchain and targeting riscv64-unknown-linux-gnu fail to execute under Spike+pk with this output:
```
$ /scratch/projects/riscv-isa-sim//build/spike \
    --varch=vlen:128,elen:64 --isa=rv64gcv_zvkned \
    /rivos/riscv-pk/riscv64-unknown-linux-gnu/bin/pk aes-cbc-test
bbl loader
bad syscall #278!
```

Syscall 278 is getrandom which is invoked prior to reaching main(). Providing a syscall stub to return ENOSYS is enough to please the caller. This has the additional benefit of not requiring to implement getrandom() in a portable way.

```
$ /scratch/projects/riscv-isa-sim//build/spike \
    --varch=vlen:128,elen:64 --isa=rv64gcv_zvkned \
    /rivos/riscv-pk/riscv64-unknown-linux-gnu/bin/pk aes-cbc-test
bbl loader
vlen: 128 bits
--- Running 'CBCKeySbox256' test suite...
...
```

### DIFF
--- a/pk/syscall.c
+++ b/pk/syscall.c
@@ -632,6 +632,7 @@ long do_syscall(long a0, long a1, long a2, long a3, long a4, long a5, unsigned l
     [SYS_rt_sigprocmask] = sys_stub_success,
     [SYS_clock_gettime] = sys_clock_gettime,
     [SYS_chdir] = sys_chdir,
+    [SYS_getrandom] = sys_stub_nosys,
   };
 
   const static void* old_syscall_table[] = {
--- a/pk/syscall.h
+++ b/pk/syscall.h
@@ -57,6 +57,7 @@
 #define SYS_set_tid_address 96
 #define SYS_set_robust_list 99
 #define SYS_madvise 233
+#define SYS_getrandom 278
 #define SYS_statx 291
 
 #define OLD_SYSCALL_THRESHOLD 1024


egouriou-rivos:
@aswaterman for your consideration. Please let me know if you want me to file an issue to document the pre-existing failure.

aswaterman:
I’m certainly OK with the stub, but I’m confused as to why it is needed. Reading the code, it appears that sys_stub_nosys is already called whenever there is no such call defined. Where is the “bad syscall” printout coming from?

egouriou-rivos:
Oh, I just didn't pay enough attention when moving from our internal version to /master .
 Rivos' internal version of riscv-pk is behind and doesn't contain ca69d69a.

```
commit ca69d69a304d5668a8c95d1de5c6c6fa71ed9936
Author: Andrew Waterman <andrew@sifive.com>
Date:   Mon Apr 11 01:22:16 2022 -0700

    Handle unimplemented syscalls gracefully
    
    We previously kernel-panicked because that made it more obvious when a
    syscall implementation was missing.  These days, it's more common that
    the C library will do something sensible in response to returning -ENOSYS.
    Favor that approach to avoid frustrating users.
```

Never mind then. I'll move our clone of riscv-pk forward. Thanks.

aswaterman:
Ah, right, I forgot about that.  Glad you figured it out.

PR #310, opened 2024-01-09
Enable shadow stack from .note.gnu.property

SuHo-llrr:
reference from llvm-project commit https://github.com/llvm/llvm-project/pull/77414

1. Parse .note.gun.property to check Zicfiss/Zicfilp properties
2. If Zicfiss on, we should allocate SS(Shadow stack) page with PROT_WRITE
3. We should allow allocate no-read page when SS page allocating

Some tasks require discussion:

- What should the default shadow stack size be?
- How to allow page allocation only for write-only pages?
> - Checking instruction with epc address? (Maybe this is more in line with hardware behavior.)
> - Checking ELF enable shstk flag?

### DIFF
--- a/pk/boot.h
+++ b/pk/boot.h
@@ -6,6 +6,7 @@
 #ifndef __ASSEMBLER__
 
 #include <stddef.h>
+#include <stdbool.h>
 
 typedef struct {
   int phent;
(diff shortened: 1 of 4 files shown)


aswaterman:
This seems like a very heavyweight addition with only minor incremental value.  I would prefer that `pk` instead just declare no support for this extension, instead unconditionally setting the various control bits so that SS/LP are disabled in this context.

cc @ved-rivos

PR #311, opened 2024-01-24
Add Zicfilp support

mylai-mtk:
This PR adds support of the mechanism defined in the [Zicfilp spec](https://github.com/riscv/riscv-cfi/blob/main/cfi_forward.adoc) (commit 079216b).
It is part of the effort to support Zicfilp in spike/pk.


pk/pk.c
```diff
@@ -13,15 +13,17 @@
...
   printk("Proxy kernel\n\n");
   printk("usage: pk [pk options] <user program> [program options]\n");
   printk("Options:\n");
-  printk("  -h, --help            Print this help message\n");
-  printk("  -p                    Disable on-demand program paging\n");
-  printk("  -s                    Print cycles upon termination\n");
+  printk("  -h, --help                    Print this help message\n");
+  printk("  -p                            Disable on-demand program paging\n");
+  printk("  -s                            Print cycles upon termination\n");
+  printk("  --zicfilp-enable-user-cfi     Enable Zicfilp CFI mechanism for user program\n");
```
aswaterman, line 26: Maybe shorten this to `--zicfilp`?

mylai-mtk:
Update: Rename `SCE_XTVAL_LANDING_PAD_FAULT` to `LANDING_PAD_FAULT` due to review in riscv-software-src/riscv-isa-sim#1582

mylai-mtk:
Update: Squash fixups, re-generate `encoding.h` from `riscv-opcodes` `master` commit

aswaterman:
Sorry for the delay.  Merged.

machine/minit.c
```diff
@@ -78,7 +78,8 @@ static void delegate_traps()
     (1U << CAUSE_BREAKPOINT) |
     (1U << CAUSE_LOAD_PAGE_FAULT) |
     (1U << CAUSE_STORE_PAGE_FAULT) |
-    (1U << CAUSE_USER_ECALL);
+    (1U << CAUSE_USER_ECALL) |
+    (1U << CAUSE_SOFTWARE_CHECK_FAULT);
```
poemonsense, line 82: What if `CAUSE_SOFTWARE_CHECK_FAULT` is not supported (never happening) in the current implementation?

In the next lines, there is an assertion check for `assert(read_csr(medeleg) == exceptions);`. This would require the `CAUSE_SOFTWARE_CHECK_FAULT` bit in medeleg to be writable. If the corresponding bit is not writable, pk will abort. Should pk have some configurable options to disable this ISA extension?

aswaterman: Yes, this is a problem, but I don't think configurability of pk is the best solution.  Simply masking off these optional codes from the assertion is the answer.  The fix is here: 4f3debe4d04f56d31089c1c716a27e2d5245e9a1

PR #318, opened 2024-03-22
Implement syscall readlinkat and readv

mylai-mtk:
I implemented them since they're used in my libc testsuite, and think they may be useful others using spike, so I'm contributing them.

`readv` is implemented by sequential `read`, since forwarding the syscall requires kernel-space dynamic memory management, which is not practically usable in pk, so the `readv` syscall is spilt into many `read`s instead.

The frontend of readlinkat is at riscv-software-src/riscv-isa-sim#1631 .


aswaterman (approved):
LGTM at a quick glance.

Issue #297, opened 2023-05-03
spike pk hello, failed with assertion failed: !(eh.e_flags & EF_RISCV_RVC)


jundijiujieke:
spike pk hello, failed with assertion failed: !(eh.e_flags & EF_RISCV_RVC)
Does any know how to fix this probelm? I don't know whether is my riscv-tool's problem or pk's problem?

bbl loader
../pk/elf.c:47: assertion failed: !(eh.e_flags & EF_RISCV_RVC)
Power off

aswaterman:
Looks like `pk` was compiled without support for the C extension, but you're trying to run a binary that requires the C extension.

jundijiujieke:
> Looks like `pk` was compiled without support for the C extension, but you're trying to run a binary that requires the C extension.

Thanks for your reply, do you mean i have to reinstall the pk with C extension.

aswaterman:
Yeah, I think building `pk` with support for the C extension will fix this issue.

jundijiujieke:
Thanks for your help, I tried your solution and fixed this problem.

aswaterman:
Great! Glad I was able to help.

PR #327, opened 2024-05-01
pk: use kassert in load_elf instead of assert

caizixian:
This avoids calling into assert and eventually `printm` (which might use UART using physical addresses, etc) in supervisor mode.

Without this patch, reproducing #297 can cause the following memory fault.
```
z  0000000000000000 ra ffffffc000006a20 sp ffffffc000427d10 gp 0000000000000000
tp 0000000000000000 t0 0000000080000004 t1 0000000080000004 t2 0000000000000000
s0 ffffffc000427d30 s1 0000000000000000 a0 000000000000002e a1 0000000000000100
a2 ffffffc000015da8 a3 0000000000000005 a4 0000000000000005 a5 0000000010000005
a6 0000000000000000 a7 0000000000000000 s2 0000000000000000 s3 0000000000000000
s4 0000000000000000 s5 0000000000000000 s6 0000000000000000 s7 0000000000000000
s8 0000000000000000 s9 0000000000000000 sA 0000000000000000 sB 0000000000000000
t3 0000000000000000 t4 0000000000000000 t5 0000000000000000 t6 0000000000000000
pc ffffffc0000086f4 va/inst 0000000010000005 sr 8000000200006100
Kernel load segfault @ 0x0000000010000005
```

### DIFF
--- a/pk/elf.c
+++ b/pk/elf.c
@@ -38,13 +38,13 @@ void load_elf(const char* fn, elf_info* info)
     goto fail;
 
 #if __riscv_xlen == 64
-  assert(IS_ELF64(eh));
+  kassert(IS_ELF64(eh));
 #else
-  assert(IS_ELF32(eh));
+  kassert(IS_ELF32(eh));
 #endif
 
 #ifndef __riscv_compressed
-  assert(!(eh.e_flags & EF_RISCV_RVC));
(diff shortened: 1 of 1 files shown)


aswaterman (approved):
Thank you.

PR #341, opened 2025-01-31
Add no-op futex syscall implementation

abejgonzalez:
This allows simple programs compiled with modern versions of GLIBC to work with `pk`. For example:

`hello.cc`:
```
#include <iostream>
int main(void) {
  std::cout << "Hello!" << std::endl;
  return 0;
}
```

Compile/test:
```
riscv64-unknown-linux-gnu-gcc hello.cc -lstdc++ -static -o hello
spike pk hello
```

Solution was previously given in:
https://github.com/riscv-software-src/riscv-pk/issues/207
https://github.com/riscv-software-src/riscv-isa-sim/issues/1443
https://github.com/riscv-software-src/riscv-isa-sim/issues/1605

### DIFF
--- a/pk/syscall.c
+++ b/pk/syscall.c
@@ -518,7 +518,7 @@ int sys_times(long* loc)
   kloc[0] = t / (CLOCK_FREQ / 1000000);
 
   memcpy_to_user(loc, kloc, sizeof(kloc));
-  
+
   return 0;
 }
 
@@ -531,7 +531,7 @@ int sys_gettimeofday(long* loc)
   kloc[1] = (t % CLOCK_FREQ) / (CLOCK_FREQ / 1000000);
 
   memcpy_to_user(loc, kloc, sizeof(kloc));
-  
+
   return 0;
 }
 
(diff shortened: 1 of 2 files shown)


pk/syscall.c
```diff
@@ -770,6 +770,7 @@ long do_syscall(long a0, long a1, long a2, long a3, long a4, long a5, unsigned l
     [SYS_readlinkat] = sys_readlinkat,
     [SYS_readv] = sys_readv,
     [SYS_riscv_hwprobe] = sys_riscv_hwprobe,
+    [SYS_futex] = sys_stub_success,
```
abejgonzalez, line 773: Re-using `sys_stub_success` instead of creating a new function.

jerryz123 (approved):
I think this is fine.... it doesn't behave exactly the same as sys_futex, but that should be ok

aswaterman (approved):
I'm going out on a limb and hoping this won't cause problems for single-threaded programs in practice.  I guess we'll see :)

